PR #48, opened 2015-01-26
Tests: Double dots in domain part, German Umlauts as valid e-mail

0x46616c6b:
Only to prevent regression.


egulias:
Great, thanks!

PR #49, opened 2015-02-03
The latest php-doctrine-lexer is needed

DavidPrevot:
The tests actually fail with version 1.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   },
   "require":      {
     "php": ">= 5.3.3",
-    "doctrine/lexer": "~1.0"
+    "doctrine/lexer": "~1.0,>=1.0.1"
   },
   "require-dev" :   {
     "satooshi/php-coveralls": "dev-master",


xabbuh:
Might be a good idea to expand the Travis config to also perform tests against the lowest allowed dependency version.

egulias:
@DavidPrevot the [tilde `~` operator](https://getcomposer.org/doc/01-basic-usage.md#next-significant-release-tilde-and-caret-operators-) defines exactly that, with the improved constraint of not allowing in this case `v2` to be fetched by error (see the example provided by composer's docs).
This is because `v2` of the lib can change the public API and behaviour of the lib, breaking this one.

@xabbuh version `1.0.1` is the lowest allowed. Tilde operator fetches that. But should be changed to `>= 1.0.1 < 2.0.0` instead to be equally to tile.
By the way, how can be configured Travis to allow for multiple dependencies versions?

xabbuh:
@egulias Well, `~1.0` does not forbid `1.0.0` which seems to be broken. Please have a look at #50 for how to test against the minimal versions. As you can see, the build with `doctrine/lexer` `1.0` fails. If we changed the dependency to be at least `1.0.1`, [the build would succeed](https://travis-ci.org/xabbuh/EmailValidator/builds/49718597).

egulias:
Thanks for the explaining PR @xabbuh , I see the flag `--prefer-lowest` is what defeats `~`.
Then I'd go with `>=1.0.1 < 2.0.0` to allow for updates within the same major version, what do you think?

DavidPrevot:
Hi,

Le 07/02/2015 19:17, Eduardo Gulias Davis a écrit :

> Then I'd go with `>=1.0.1 < 2.0.0` to allow for updates within the same major version, what do you think?

That would be fine and equivalent to the proposed `~1.0,>=1.0.1`, thanks.

Regards

David

xabbuh:
>  Then I'd go with >=1.0.1 < 2.0.0 to allow for updates within the same major version, what do you think?

Afaik, it's just a matter of taste what you want to use. You could also choose `^1.0.1` if you preferred that.

egulias:
That's true @DavidPrevot , we can go with this PR.
Thanks both!

PR #58, opened 2015-04-23
update Travis configuration

xabbuh:
* run builds on the container-based infrastructure
* add a build for PHP 7
* report the build status as soon as possible (either when a mandatory
  job fails or when only jobs allowed to fail are left)


egulias:
Thanks!

PR #75, opened 2015-06-26
update installation instructions

xabbuh:
Given that Composer will install the latest version of a package
matching the stability restrictions that apply to project, users can
get an undesired version of package when they don't specify a version
constraint.


egulias:
Good point. Thanks

Issue #80, opened 2015-10-01
Email considered valid with a closing parenthesis


whyte624:
Example:
$email = "test@example.com)";
Validation gives true. Expected to be false.

PR #81, opened 2015-10-02
#80 validate number of closing parenthesis

whyte624:
Fix issue #80.
Checking number of closing parenthesis and compare it with number of parenthesis opened before.

### DIFF
--- a/src/Egulias/EmailValidator/Parser/Parser.php
+++ b/src/Egulias/EmailValidator/Parser/Parser.php
@@ -9,6 +9,7 @@ abstract class Parser
 {
     protected $warnings = array();
     protected $lexer;
+    protected $openedParenthesis = 0;
 
     public function __construct(EmailLexer $lexer)
     {
@@ -21,6 +22,12 @@ abstract class Parser
     }
 
     abstract public function parse($str);
+
+    /** @return int */
+    public function getOpenedParenthesis()
+    {
+        return $this->openedParenthesis;
+    }
 
     /**
      * validateQuotedPair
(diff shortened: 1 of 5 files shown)


src/Egulias/EmailValidator/Parser/Parser.php
```diff
@@ -36,14 +36,18 @@ protected function validateQuotedPair()
     }
 
     /**
-     * @return string the the comment
-     * @throws \InvalidArgumentException
+     * @param int $openedParenthesis
+     * @return string the comment
      */
-    protected function parseComments()
+    protected function parseComments(&$openedParenthesis = 0)
```
egulias: Instead of passing by reference (which causes hidden side effects), the method can return the number of parenthesis found. What do you think?

whyte624: I was thinking of it, but it looked not obvious for method parseComments to return number of parenthesis.

egulias: And making it protected within `Parser` class? We can initialize it to 0 outside the loop of the `Local` or `Domain` parsers.
Similar side-efectish but more OOP.
WDYT?

whyte624: Do you mean to change the $openParenthesis variable to a protected field?

egulias: Yes.

whyte624:
If it's alright now, let's proceed with a release. I would like to have those changes in my project. Thanks!

egulias:
Sorry for the delay.

egulias:
1.2.10 is out with your changes.

whyte624:
Great! Thanks!

Issue #87, opened 2016-02-15
Email not verified properly


zeeshankhan81:
Steps to Reproduce:
1. Create a content type having email as one of the field.
2. Enter a invalid email such as "example@example".

It will validate it but it is not a valid email address.

PR https://github.com/egulias/EmailValidator/pull/88

egulias:
Hi @zeeshankhan81 
"example@example" is a valid email.

PR #88, opened 2016-02-15
Fixed email verification issue.

zeeshankhan81:
Refer Drupal 8 issues: https://www.drupal.org/node/2661278
Refer git issue: https://github.com/egulias/EmailValidator/issues/87

### DIFF
--- a/src/Egulias/EmailValidator/Parser/DomainPart.php
+++ b/src/Egulias/EmailValidator/Parser/DomainPart.php
@@ -31,6 +31,10 @@ class DomainPart extends Parser
             $this->parseDomainComments();
         }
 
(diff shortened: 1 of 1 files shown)


dawehner:
We should expand the test coverage here ...

egulias:
Tahnks @zeeshankhan81 for the PR. But:
Won't merge. Two reasons:

1. "email@example" is a valid email, please see the [tests](https://github.com/egulias/EmailValidator/blob/1.2/tests/egulias/Tests/EmailValidator/EmailValidatorTest.php#L44) and [RFC5321](http://www.rfcreader.com/#rfc5321_line550)
2. Instead of a regexp you should use the provided tokens (I'd have commented in code if it not where because of 1)

PR #89, opened 2016-03-15
Adding documentation and consolidating some code for readability

natechicago:
I've added documentation to this class to make it easier for others to understand (at a quick glance) --
 - the differences between errors and warnings
 - the different categories of warnings
 - the exact RFCs that this class validates against (please confirm this list)

Additionally, some simple refactorings were applied to make a few parts of the code quicker/easier to read.  They're logically equivalent to the original code, and the tests still pass.

### DIFF
--- a/src/Egulias/EmailValidator/EmailValidator.php
+++ b/src/Egulias/EmailValidator/EmailValidator.php
@@ -9,48 +9,50 @@ namespace Egulias\EmailValidator;
  */
 class EmailValidator
 {
+    /**
+     * Critical validation errors used to indicate that
+     * an email address is invalid:
+     */
     const ERR_CONSECUTIVEATS     = 128;
     const ERR_EXPECTING_DTEXT    = 129;
     const ERR_NOLOCALPART        = 130;
     const ERR_NODOMAIN           = 131;
     const ERR_CONSECUTIVEDOTS    = 132;
     const ERR_ATEXT_AFTER_CFWS   = 133;
-    const ERR_ATEXT_AFTER_QS     = 134;
-    const ERR_ATEXT_AFTER_DOMLIT = 135;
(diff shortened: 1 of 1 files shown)


src/Egulias/EmailValidator/EmailValidator.php
```diff
@@ -74,6 +112,26 @@ public function __construct()
...
 
+    /**
+     * Validates an email address against the following standards:
+     *
+     * RFC-5322: Internet Message Format
+     * RFC-5321: Simple Mail Transfer Protocol
+     * RFC-6530: Overview and Framework for Internationalized Email
+     * RFC-1123 section 2.1: Requirements for Internet Hosts -- Application and Support
+     * RFC-4291 section 2.2: IP Version 6 Addressing Architecture
+     *
```
egulias: Please add

- [6530](http://tools.ietf.org/html/rfc6530#section-4.6)
- [6531](http://tools.ietf.org/html/rfc6531)
- [6532](http://tools.ietf.org/html/rfc6532) 

which are for **SMTPUTF8** international emails.

egulias:
Thanks @natechicago 
I added a comment about the list of supported RFCs.
I'll merge when updated.

src/Egulias/EmailValidator/EmailValidator.php
natechicago: Okay, will do.  In my mind, I had been treating 6530 as the 'parent' to 6531 and 6532, but I suppose that's not really a fair way of looking at those RFCs!

natechicago:
Thanks for the quick review, and for offering your feedback over in Symfony's 18156.  I'm also in favor of 'proposal 1', so I hope the other commenters will find it acceptable.  I'm going to hold off on making any changes in the corresponding PR until they have a chance to chime in.

Issue #101, opened 2016-04-30
Proposal: 2.0 API


ro0NL:
Hi,

I forked your lib today as I was planning to investigate it for #93. I think before adding new features, some core components need to be improved first, api-wise, and imho of course ;-) I saw you already did some work on errors & warnings. I'm talking `master` here.
## The `Lexer`

I don't want to discuss any implementation/spec details here, im not an email expert nor a lexer expert. As its API is provided by doctrine we should keep this as is.
## The `Parser`

This one is a bit weird i think, basicly there are 3 parsers, however api- and implementation-wise they are all a bit different..
### `EmailParser`
- Domain part is set using `EmailParser::$domainParser`, i.e. using the lexer, which is good.
- Local part is set using simple [`explode`](https://github.com/egulias/EmailValidator/blob/master/src/Egulias/EmailValidator/EmailParser.php#L80) which somehow bypasses this [test](https://github.com/egulias/EmailValidator/blob/master/tests/egulias/Tests/EmailValidator/Validation/RFCValidationTest.php#L87), wtf?
- Returns parsed parts which is good.
### `LocalPart`
- Doesn't use argument `$localPart` in `parse()`, it relies on `EmailParser` calling `setInput()` on the same lexer instance previously.
- Returns no parsed part(s) nor does it allow for `getLocalPart()`, the logic is done back in `EmailParser::parse()` but that is wrong for using explode.. im lost here.
### `DomainPart`
- Same flaw as `LocalPart` (argument `$domainPart` is unused in `parse()`).
- Returns no parsed part(s) but allows for `getDomainPart()`.
### Proposal

I think the problem is `LocalPart` and `DomainPart` are not parsers itself, but they modify the `Lexer` state, to make the actual parser, i.e. `EmailParser`, work. Then again `DomainPart` _does_ work as a parser. Point is.. i really think we should use the lexer/local part parser for getting the "correct" local part, i just dont get this explode thingy. I guess the individual parts as result from the parser are just unused/untested. This should work! So...;
- Make `LocalPart` a parser like `DomainPart`, i.e. allow for `getLocalPart()` and use it in `EmailParser`
  - Point is.. if we have standalone parsers, they should work standalone. I.e. all setting its input on the lexer.
- Move all the logic to `EmailParser`, i.e. perhaps it's a bit over-engineered
- Make it part of the lexer component, i.e. just modify state on the lexer and grab the parts from it in the parser
[...]

Issue #68, opened 2015-05-04
Make Parsers start from "scratch"


egulias:
Currently localPart and domainPart parsers depend on the position where the lexer is when they start parsing.
They should 'lex' the string that they get, the caller is the responsible for knowing what is passing.
Check https://github.com/egulias/EmailValidator4J

PR #102, opened 2016-05-01
Allow for escaped quote in quoted text

ro0NL:
According to https://en.wikipedia.org/wiki/Email_address#Local_part we should expect an email address like `"very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com`.

I.e. an escaped quote in a quoted text part, currently the local part parser reaches until `"very.(),:;<>[]\`.


ro0NL:
FYI i think the right solution for #101 and #68 is to let the parser return a full AST for the given email address, from which you can reconstruct the email adress exactly as given. I.e. it allows for formatters. That way you would also have warnings _per_ part as currently domain warnings can override local warnings, which is fine from a validation POV but we lose context.

egulias:
It probably is. Lets chat on #101

PR #115, opened 2016-06-20
fixed DNSCheckValidation bug

issei-m:
The current `DNSCheckValidation` does not allow the email containing `+`. This is a problem as the Gmail users sometimes use `+` in local-part (e.g. `issei.m7+foo@gmail.com`) to generate multiple account.


egulias:
Thanks!

PR #117, opened 2016-06-23
fixed type missing and removed unused code

issei-m:
To be more friendly to IDE 😸


egulias:
Not that intrusive ;)

PR #119, opened 2016-06-24
Add an EmailValidatorInterface

cilefen:
This will make it easier to [replace implementations](https://www.drupal.org/node/2749873).


egulias:
Hi @cilefen 
Would be a great addition. 
You might have noticed that the lib got a new major version, v2.
v2 comes with interfaces for email validation strategies (see https://github.com/egulias/EmailValidator#how-to-extend). That would make super easy to interchange validations.

A question from my side: would it be possible for Drupal 8 to use v2 of the lib?

Would you mind pointing the PR to the 1.2 branch? Being a non braking change I could merge it without the need for a major version.

Thanks!

cilefen:
#120

PR #120, opened 2016-06-25
Add EmailValidatorInterface in 1.x

cilefen:
This will make it easier to [replace implementations](https://www.drupal.org/node/2749873).


cilefen:
I opened an [issue](https://www.drupal.org/node/2755401) for Drupal to upgrade to 2.

cilefen:
We will want to do this in 2.x as well.

PR #122, opened 2016-07-05
MultipleValidatorWithAnd now can break out of loop when error occurs

issei-m:
I often use `MultipleValidatorWithAnd` to combine validations such as `RFCValidation`, `DNSCheckValidation`.
In this case, when the preceding RFC check fails, I wanna skip the succeeding DNS check process since that would be costly.

That's why I would like to add a single option for this class to break out of validation loop when one of validations gets failure.

### DIFF
--- a/EmailValidator/Validation/MultipleValidationWithAnd.php
+++ b/EmailValidator/Validation/MultipleValidationWithAnd.php
@@ -3,23 +3,49 @@
 namespace Egulias\EmailValidator\Validation;
 
 use Egulias\EmailValidator\EmailLexer;
+use Egulias\EmailValidator\Exception\InvalidEmail;
 use Egulias\EmailValidator\Validation\Exception\EmptyValidationList;
 
 class MultipleValidationWithAnd implements EmailValidation
 {
+    /**
+     * @var EmailValidation[]
+     */
     private $validations = [];
+
+    /**
+     * @var array
+     */
     private $warnings = [];
+
+    /**
+     * @var InvalidEmail
+     */
     private $error;
-    
-    public function __construct(array $validations)
+
+    /**
+     * @var bool
+     */
+    private $breakIfError;
+
+    /**
+     * @param EmailValidation[] $validations  The validations.
+     * @param bool              $breakIfError If true, it breaks out of validation loop when error occurs,
+     *                                        it means returned MultipleErrors might not contain all causes of errors. (false by default)
+     */
+    public function __construct(array $validations, $breakIfError = false)
     {
         if (count($validations) == 0) {
             throw new EmptyValidationList();
         }
         
         $this->validations = $validations;
+        $this->breakIfError = $breakIfError;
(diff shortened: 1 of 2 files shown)


egulias:
How about adding a constant as a flag so the API has more meaning. E.g
 ```php
const STOP_ON_ERROR = true;
const ALLOW_ALL_ERRORS = false;

public function __construc(array $validatons, $breakCondition = ALLOW_ALL_ERRORS)
```

File: EmailValidator/Validation/MultipleValidationWithAnd.php [hunk 1]
```diff
@@ -3,23 +3,49 @@
...
+
+    /**
+     * @var bool
+     */
+    private $breakIfError;
+
+    /**
+     * @param EmailValidation[] $validations  The validations.
+     * @param bool              $breakIfError If true, it breaks out of validation loop when error occurs,
+     *                                        it means returned MultipleErrors might not contain all causes of errors. (false by default)
```
egulias: `it means returned MultipleErrors will contain a single error, the first found.`

EmailValidator/Validation/MultipleValidationWithAnd.php
```diff
@@ -29,17 +55,27 @@ public function isValid($email, EmailLexer $emailLexer)
             $result = $result && $validation->isValid($email, $emailLexer);
             $this->warnings = array_merge($this->warnings, $validation->getWarnings());
             $errors[] = $validation->getError();
+
+            if (!$result && $this->breakIfError) {
```
egulias: WDYT of encapsulating into a method? Like
```php
if($this->shouldStop($result)) {
    break;
}
```

EmailValidator/Validation/MultipleValidationWithAnd.php
```diff
@@ -3,23 +3,49 @@
 namespace Egulias\EmailValidator\Validation;
 
 use Egulias\EmailValidator\EmailLexer;
+use Egulias\EmailValidator\Exception\InvalidEmail;
 use Egulias\EmailValidator\Validation\Exception\EmptyValidationList;
 
 class MultipleValidationWithAnd implements EmailValidation
 {
+    /**
+     * @var EmailValidation[]
```
egulias: Not sure here. Shouldn't this be
`@var array EmailValidation`?

issei-m:
@egulias Updated now to change following:

- renamed `$breakIfError` option to `$mode`
- put constants for $mode
- extracted skip logic into private method

EmailValidator/Validation/MultipleValidationWithAnd.php
issei-m: Intends to make IDE's analyzer smarter, each elements of `$validations` now can be guessed its type in the loop block like this:

image

egulias:
I'll issue a new release next weekend, thanks!

PR #124, opened 2016-07-21
MultipleValidationWithAnd::getError() now returns NULL if no errors

issei-m:
Hi, I'm about to take this topic for three times :)
However MultipleValidationWithAnd is still wrong I think.

EmailValidation::getError() intends to return the NULL if no errors, but this class won't do so.

### DIFF
--- a/EmailValidator/Validation/EmailValidation.php
+++ b/EmailValidator/Validation/EmailValidation.php
@@ -9,7 +9,7 @@ use Egulias\EmailValidator\Warning\Warning;
 interface EmailValidation
 {
     /**
-     * Returns true if the given email is fine.
+     * Returns true if the given email is valid.
      *
      * @param string     $email      The email you want to validate.
      * @param EmailLexer $emailLexer The email lexer.
@@ -20,6 +20,8 @@ interface EmailValidation
 
     /**
      * Returns the validation error.
+     *
+     * Be careful about no-error does not necessarily mean email was valid.
      *
      * @return InvalidEmail|null
      */
--- a/EmailValidator/Validation/EmailValidation.php
+++ b/EmailValidator/Validation/EmailValidation.php
@@ -21,8 +21,6 @@ interface EmailValidation
     /**
      * Returns the validation error.
      *
-     * Be careful about no-error does not necessarily mean email was valid.
-     *
      * @return InvalidEmail|null
      */
     public function getError();
--- a/EmailValidator/Validation/EmailValidation.php
+++ b/EmailValidator/Validation/EmailValidation.php
@@ -4,18 +4,31 @@ namespace Egulias\EmailValidator\Validation;
 
 use Egulias\EmailValidator\EmailLexer;
 use Egulias\EmailValidator\Exception\InvalidEmail;
+use Egulias\EmailValidator\Warning\Warning;
 
 interface EmailValidation
 {
+    /**
+     * Returns true if the given email is fine.
+     *
+     * @param string     $email      The email you want to validate.
+     * @param EmailLexer $emailLexer The email lexer.
+     *
+     * @return bool
+     */
     public function isValid($email, EmailLexer $emailLexer);
 
     /**
-     * @return InvalidEmail
+     * Returns the validation error.
+     *
+     * @return InvalidEmail|null
      */
     public function getError();
 
     /**
-     * @return array of Warning
+     * Returns the validation warnings.
+     *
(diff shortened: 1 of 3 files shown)


EmailValidator/Validation/MultipleValidationWithAnd.php
```diff
@@ -71,8 +70,11 @@ public function isValid($email, EmailLexer $emailLexer)
                 break;
             }
         }
-        $this->error = new MultipleErrors($errors);
-        
+
+        if (!empty(array_filter($errors))) {
```
egulias: We can instead filter them before adding them to `$errors`, replacing `$errors[] = $validation->getError();` with something like this:
```php
//remove $errors = [];
//...
foreach ($this->validations as $validation) {
    //...
    $errors = $this->addNewError($validation->getError(), $errors);
    // ...
}
//...
private function addNewError($possibleError, array $errors) {
    if(null !== $possibleError) {
        $errors[] = $possibleError;
    }
    return $errors;
}
```
I propose this way because I'm not a fan of passing by reference, so I prefer a function that returns the new array.
Also the error list is being augmented only when necessary.

What do you think?

egulias:
Hi! Thanks for your third improvement :)

issei-m:
@egulias Sorry for late reply. I have updated to fix mentioned point.

EmailValidator/Validation/EmailValidation.php
```diff
@@ -21,6 +21,8 @@ public function isValid($email, EmailLexer $emailLexer);
     /**
      * Returns the validation error.
      *
+     * Be careful about no-error does not necessarily mean email was valid.
```
issei-m: @egulias This is correct, right?

EmailValidator/Validation/MultipleValidationWithAnd.php
```diff
@@ -64,20 +64,28 @@ public function isValid($email, EmailLexer $emailLexer)
...
 
-        if (!empty(array_filter($errors))) {
+        if (!empty($errors)) {
             $this->error = new MultipleErrors($errors);
         }
 
         return $result;
     }
 
+    private function addNewError($possibleError, array $errors) {
```
egulias: Please move `{` to a new line.

EmailValidator/Validation/EmailValidation.php
egulias: Well, for any given EmailValidation implementation, having no errors means it is valid in terms of that validator. What could be a case for an invalid email without error?

issei-m: `NoRFCWarningsValidation` [can be](https://github.com/issei-m/EmailValidator/blob/68aae522c42d8253f415480e11ba8938e984b343/EmailValidator/Validation/NoRFCWarningsValidation.php#L11-L11).
This class will have no error if RFC check is success but with warnings.

issei-m: I'm assuming `NoRFCWarningsValidation` seems to be on wrong side. 
So I will concern about this in another PR. (I'm gonna remove above line here.)

Tests/EmailValidator/Validation/MultipleValidationWitAndTest.php
```diff
@@ -32,6 +32,19 @@ public function testEmptyListIsNotAllowed()
         new MultipleValidationWithAnd([]);
     }
 
+    public function testValidationIsValid()
+    {
+        $lexer = $this->getMock("Egulias\\EmailValidator\\EmailLexer");
```
henriquemoody: You could use `EmailLexer::class`

issei-m: Exactly. but I'm gonna use existing CS in here.
However, `::class` should be used in whole of PJ because we are using PHP 5.5 or later.

issei-m:
@egulias updated.

btw, Should [this](https://github.com/egulias/EmailValidator/pull/124#discussion_r72366596) be a bug, I will work to fix.

EmailValidator/Validation/EmailValidation.php
egulias: That's great, thanks!

egulias:
It is an error in fact. It should return an error stating the fact that there were warnings.

issei-m:
@egulias OK, I'm gonna take this up in another PR. So could this PR get merged? ;)

PR #127, opened 2016-08-15
Update composer.json to use the right extension string

bradchesney79:
By using
"ext-intl": "*"
in the require block, the suggestion will be hidden if it is already available. ext/php-intl has few google search results and will confuse less knowledgeable users.

https://github.com/egulias/EmailValidator/issues/128


egulias:
Thanks!

PR #137, opened 2017-01-25
Compatibility with DNS RFC 1034

LeShadow:
In one of my commits, there is a wrong RFC specification, this should be RFC 1034.

According to RFC 1034, every DNS query needs to have a . at the end of
the name record you are requesting.

This will show that that name record is the root of the domain you are
asking
In some cases, when the . is missing, the hostname of the server the
code is running on will be appended to the name record when
checkdnsrr() is being executed and this will in turn return a false
positive when there is a wildcard DNS record for the root domain of the
hostname of the server.

The first commit adds the . to the hostname when you do a dns query with checkdnsrr()
The second commit fixes a few deprecation warnings in the tests


egulias:
Hi @LeShadow 
Which was the warning you were getting?
Another PR (#136) has brought conflicts with yours. Do you mind fixing them with a rebase (to have a clean history?).
Thanks!

egulias:
You can also take the chance to amend the wrong RFC number on the commit message.

LeShadow:
@egulias My apologies for the mess. My experience with git is limited and I haven't done much with it.
The PR should be good now. Although I tried to edit my commit message and somehow failed.

I didn't really get a warning, I found out that this library didn't followed the RFC because of a code my client was executing.

He was using Drupal 8.x, which uses this library for email validation. Every time he checked an email and he didn't add a . to the hostname (name record), the server hostname's root domain was added to the request.

This is how I found out that you should add a . to the hostname

nielsaers:
If we want to fix this for Drupal 8 we'd have to fix version 1.2.* Currently Drupal 8 still uses version 1.2 (https://github.com/drupal/drupal/blob/8.3.x/core/composer.json)

I don't know what your policy is of applying fixes to older versions?

egulias:
Hi @nielsaers 
I'm keeping 1.2.* as much as possible, however porting fixes require a work explicitly for 1.2 as it is not compatible with v2.
I've created the issue https://github.com/egulias/EmailValidator/issues/138 to port it. PR's are welcome!

egulias:
Hey @LeShadow don't be so harsh to yourself :). I'm happy to help you here.
For some reason you just brought all the other change without respecting the original commits and it is now looking that you are bringing a lot of changes.

As @nielsaers pointed out, these should be done in branch 1.2 as well.

Either way, I'll try fix the PR. I might need to do a new one on your behalf.

egulias:
See https://github.com/egulias/EmailValidator/pull/139

PR #155, opened 2017-11-11
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06), that support this namespace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ install:
 
 script:
  - mkdir -p build/logs
- - phpunit --coverage-clover build/logs/clover.xml
+ - vendor/bin/phpunit --coverage-clover build/logs/clover.xml
 
 after_script:
  - php vendor/bin/coveralls
-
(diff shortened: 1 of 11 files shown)


egulias:
Hi @gabriel-caruso,
Coveralls is failling with HHVM due to a non fixed version on `coverall` dependency. By upgrading the lock you've moved up the version. This seems mostly harmless except for [this change in config files](https://github.com/php-coveralls/php-coveralls/blob/50c60bb64054974f8ed7540ae6e75fd7981a5fd3/CHANGELOG.md#L12).

Do you mind fixing `coveralls` to `^1.0.*` and upload a new `.lock` file?

Thanks.

gabriel-caruso:
@egulias Okay, I’ll update it and run the tests again

gabriel-caruso:
@egulias Done. I've also updated `Travis CI ` to use the installed `PHPUnit` version, not the global one.

PR #159, opened 2017-12-06
Use assertCount and assertArrayHasKey

carusogabriel:


egulias:
Hi @carusogabriel , could you explain the rationale behind the change? Thanks.

carusogabriel:
Yes. The message if this test fails will be `Failed asserting that actual size x matches expected size y`, and I guess is better for debugging.

PR #166, opened 2018-04-09
DNS check should convert unicode domains first

lode:
I found this comment on php.net http://php.net/manual/en/function.checkdnsrr.php#113537:

> Before you check domain, you must convert to ascii with idn_to_ascii function:
http://us2.php.net/manual/en/function.idn-to-ascii.php .

```
var_dump(checkdnsrr('ñandu.cl', 'A')); // returns false
var_dump(checkdnsrr(idn_to_ascii('ñandu.cl'), 'A')); // return true
```


egulias:
Thanks @lode

PR #167, opened 2018-04-09
Prevent some DNS checks when not allowing local domains

lode:
I find quite some of the failed validations I have in production are people just forgetting a dot in the domain part. As I don't allow local domains for the email address this helps in not having to make a more expensive DNS check with a long timeout.

### DIFF
--- a/EmailValidator/Validation/NonLocalValidation.php
+++ b/EmailValidator/Validation/NonLocalValidation.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Egulias\EmailValidator\Validation;
+
+use Egulias\EmailValidator\EmailLexer;
+use Egulias\EmailValidator\Exception\InvalidEmail;
+use Egulias\EmailValidator\Validation\Error\LocalEmail;
+use \Spoofchecker;
+
+class NonLocalValidation implements EmailValidation
+{
+    /**
+     * @var InvalidEmail
+     */
+    private $error;
+
+    public function __construct()
+    {
+        if (!class_exists(Spoofchecker::class)) {
+            throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
+        }
+    }
+
+    public function isValid($email, EmailLexer $emailLexer)
+    {
+        // use the input to check DNS if we cannot extract something similar to a domain
+        $host = $email;
+
+        // Arguable pattern to extract the domain. Not aiming to validate the domain nor the email
+        if (false !== $lastAtPos = strrpos($email, '@')) {
+            $host = substr($email, $lastAtPos + 1);
+        }
+
+        if (strpos($host, '.') === false) {
+            $this->error = new LocalEmail();
+        }
+
+        return $this->error === null;
+    }
+
+    public function getError()
+    {
+        return $this->error;
+    }
+
+    public function getWarnings()
+    {
+        return [];
+    }
+}
--- a/EmailValidator/Validation/NonLocalValidation.php
+++ b/EmailValidator/Validation/NonLocalValidation.php
@@ -5,7 +5,6 @@ namespace Egulias\EmailValidator\Validation;
 use Egulias\EmailValidator\EmailLexer;
 use Egulias\EmailValidator\Exception\InvalidEmail;
 use Egulias\EmailValidator\Validation\Error\LocalEmail;
-use \Spoofchecker;
 
 class NonLocalValidation implements EmailValidation
 {
@@ -13,13 +12,6 @@ class NonLocalValidation implements EmailValidation
      * @var InvalidEmail
      */
     private $error;
-
-    public function __construct()
-    {
-        if (!class_exists(Spoofchecker::class)) {
-            throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
-        }
-    }
 
     public function isValid($email, EmailLexer $emailLexer)
     {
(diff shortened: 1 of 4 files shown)


issei-m (changes requested):

EmailValidator/Validation/NonLocalValidation.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @var InvalidEmail
+     */
+    private $error;
+
+    public function __construct()
+    {
+        if (!class_exists(Spoofchecker::class)) {
+            throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
+        }
```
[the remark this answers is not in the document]
lode: :eyes: copy-pasta

EmailValidator/Validation/NonLocalValidation.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        // use the input to check DNS if we cannot extract something similar to a domain
+        $host = $email;
+
+        // Arguable pattern to extract the domain. Not aiming to validate the domain nor the email
+        if (false !== $lastAtPos = strrpos($email, '@')) {
+            $host = substr($email, $lastAtPos + 1);
+        }
+
+        if (strpos($host, '.') === false) {
```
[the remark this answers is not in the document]
lode: Yeah I didn't really know what to do as other parts of the code mix both styles. As I copied the first part from another validator and personally prefer not using yoda I did this without. But that is indeed messy, didn't spot that. I'll adjusted it to use your convention!

lode:
The failing test is weird, it complains that I'm not using the `$emailLexer`, but that is part of the interface. I could let it default to `null` of course? Don't know if that helps.

egulias:
Hi @lode , just ignored the issue. 
Regarding the functionality, I understand your rationale but I'm giving it a thought, not sure if is something that it should be as a functionality. And if so, what do you think of the HTML5 email regex? It has a broader application and fulfills your need.

lode:
Thanks for your thoughts. The HTML5 validation is a good one, I don't really know what happens in these cases, because I _do_ use the input type. Maybe it is their browser that doesn't support it, or allows to skip or disable that feature?

Apart from that, I think validation is always the responsibility of the backend. It is not necessary an HTML5 client that connects to my API.

Maybe there's also another solution possible. The main issue I bumped into is that the DNS check just takes around 20 seconds before it gives a timeout on wrong domains. And there doesn't seem to be a way to adjust that. I'm okay waiting a little bit, but this is quite much before allowing an address to be submitted.

Tests/EmailValidator/Validation/NonLocalValidationTest.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * @dataProvider invalidEmailsProvider
+     */
+    public function testInvalidNonLocal($invalidEmail)
+    {
+        $validation = new NonLocalValidation();
+        $this->assertFalse($validation->isValid($invalidEmail, new EmailLexer()));
+        
+        $expectedError = new LocalEmail();
+        $this->assertEquals($expectedError, $validation->getError());
```
peter279k: This assertion I think is not correct to assert the expected error message is same.
I think it should assert the two error messages are same.

joelharkes: assertEquals checks inf content (property values) equals instead of same object reference i think?
issame checks for same object reference, so in that case this should be fine? 

(i would do an instance of check, since thats what you really want i think?)

joelharkes (changes requested):
check insteanceof instead of assertEquals.

Tests/EmailValidator/Validation/NonLocalValidationTest.php
peter279k: I think it should use the `assertInstanceOf` to check whether the class instance is `LocalEmail` class here, not using `assertEquals`.

peter279k: After tracing `NonLocalValidation` class, the `$this->error` will be the bool value or `LocalEmail` class instance.

It should use the different assertion for asserting the different `$this->error` value.
Consider following code:

```php

$this->assertInstanceOf(LocalEmail::class, $validation->getError());

$this->assertTrue($validation->getError());

$this->assertFalse($validation->getError());
```

lode: Very true, fixed the instanceof check.

peter279k:
@lode, the latest Travis CI build is broken because the `Composer` checks the peer fingerprint is failed.

Do you have any idea about fixing this problem?

peter279k:
After checking the `Composer` [source code](https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L354), it looks like this exception message will happen in `php<5.6`.

lode:
Nope. Only thing I found is that it is indeed a php 5.5 specific emulation which fails: https://stackoverflow.com/questions/47527455/getting-an-error-peer-fingerprint-did-not-match-after-running-composer-update

Further, my branch is probably not up-to-date with master anymore. That might help as well. (Should I merge or rebase?) As the error doesn't seem to be related to anything in this PR.

peter279k:
I think using rebasing commits is the one of proper way to fix this.

lode:
Hmm, that didn't seem to have fixed the issue. Although there is one less php version having this problem. It seems to me more like an issue unrelated to this PRs code.

egulias:
A bit late on this, Thanks for your patience. 
As said, is no core to the library this functionality for v2.
If you still wish to add it, please do it to V3, within the "extra" folder (please follo https://github.com/egulias/EmailValidator/blob/3.x/CONTRIBUTING.md) .

PR #172, opened 2018-06-26
[1.x] Fix result of isValid when DNS lookup fails and strict is false 

mlocati:
Let's assume that:
- `$email` contains a valid email address, but for a domain that does not exist (that is, no MX record is found)
- we call `isValid($email, true, false)`

I'd expect that the result is `false` because I specified `true` for `$checkDNS`. BTW the result is `true`.


egulias:
Hi @mlocati , thanks for the PR. It is correct, dns check it tied to an strict (no warnings) validation.
Please correct the failing tests (would be very kind from you if also fixing Travis for 5.3)

Thanks!

.travis.yml
```diff
@@ -2,12 +2,17 @@ sudo: false
 
 language: php
 
+cache:
+  directories:
+    - $HOME/.composer/cache
```
xabbuh: I would only cache `$HOME/.composer/cache/files` here. Metadata are likely to change far more often.

egulias:
Sorry for my very late response @mlocati ! Thaks for your work, will merge and release a new minor version.

PR #173, opened 2018-08-13
INTL_IDNA_VARIANT_UTS46 is not supported on all systems

tomsommer:
Fall back to INTL_IDNA_VARIANT_2003 if INTL_IDNA_VARIANT_UTS46 is not found.

### DIFF
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -44,7 +44,13 @@ class DNSCheckValidation implements EmailValidation
 
     protected function checkDNS($host)
     {
-        $host = rtrim(idn_to_ascii($host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), '.') . '.';
+        if ( defined('INTL_IDNA_VARIANT_UTS46') ) {
+            $variant = INTL_IDNA_VARIANT_UTS46
+        } else {
+            $variant = INTL_IDNA_VARIANT_2003;
+        }
+        
+        $host = rtrim(idn_to_ascii($host, IDNA_DEFAULT, $variant), '.') . '.';
 
         $Aresult = true;
         $MXresult = checkdnsrr($host, 'MX');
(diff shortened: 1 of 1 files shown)


EmailValidator/Validation/DNSCheckValidation.php
```diff
@@ -44,7 +44,13 @@ public function getWarnings()
 
     protected function checkDNS($host)
     {
-        $host = rtrim(idn_to_ascii($host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), '.') . '.';
+        if ( defined('INTL_IDNA_VARIANT_UTS46') ) {
+            $variant = INTL_IDNA_VARIANT_UTS46
```
egulias: You are missing a `;`

EmailValidator/Validation/DNSCheckValidation.php
```diff
@@ -44,7 +44,13 @@ public function getWarnings()
 
     protected function checkDNS($host)
     {
-        $host = rtrim(idn_to_ascii($host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), '.') . '.';
+        if ( defined('INTL_IDNA_VARIANT_UTS46') ) {
+            $variant = INTL_IDNA_VARIANT_UTS46;
+        } else {
```
egulias: Instead of using `else` you can initialize `$variant` with the default value `INTL_IDNA_VARIANT_2003` and only leave the `if` block.

tomsommer: I know, but `INTL_IDNA_VARIANT_2003` is deprecated, so it's an odd default value to have.

egulias: The current code uses it as a default anyway. Also it's been recently deprecated (startingPHP 7.2) so might be more sensible default than `INTL_IDNA_VARIANT_UTS46`.

egulias: @tomsommer  please change it to remove the `else` block so I can merge and create a release with all the other changes you have submitted.
Thanks!

PR #175, opened 2018-08-13
DNSCheckValidation requires the Intl extension

tomsommer:
DNSCheckValidation requires the Intl extension.

This adds the check and documents the requirement

### DIFF
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -18,6 +18,13 @@ class DNSCheckValidation implements EmailValidation
      * @var InvalidEmail
      */
     private $error;
+    
+    public function __construct()
+    {
+        if (!extension_loaded('intl')) {
+            throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
+        }
+    }
 
     public function isValid($email, EmailLexer $emailLexer)
     {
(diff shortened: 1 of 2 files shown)


EmailValidator/Validation/DNSCheckValidation.php
```diff
@@ -18,6 +18,13 @@ class DNSCheckValidation implements EmailValidation
      * @var InvalidEmail
      */
     private $error;
+    
+    public function __construct()
+    {
+        if (!extension_loaded('intl')) {
+            throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
```
egulias: Instead of a generic message, what about something like `%s feature requires the Intl extension.` ?

tomsommer: I just copied the same message from the Spoof validator :)

ohartl:
Could it be an option to use a polyfill like https://github.com/symfony/polyfill-intl-icu, so when someone cannot install ext-intl on their system it still works? We had the problem after upgrading from 2.1.4 to 2.1.5.

For example the package `symfony/intl` does the same, see https://github.com/symfony/intl/blob/master/composer.json

@egulias @tomsommer Whats your opinion on that?

tomsommer:
I would put the polyfill in `suggest` at maximum, `ext-intl` is rather common these days.

tomsommer:
I know, but having it as a dependency seems like bloat to me - that's just my two cents :)

egulias:
Thanks for the suggestion @ohartl . Do you mind opening an issue with the suggestion? I'd like to take a look at it before making a decission and here hidden in a merged PR might get lost.

Thanks!

PR #178, opened 2018-09-24
Added autoload-dev

samnela:
- Added autoload-dev in composer.json to define autoload rules for development purposes. See https://getcomposer.org/doc/04-schema.md#autoload-dev


egulias:
Thanks.

PR #179, opened 2018-10-05
Update MultipleValidationWithAnd.php

joelharkes:
Hey was debugging my code and i found this issue (not that i need it but i still fixed it anyways).

Fix all validations actually get executed when mode is ALLOW_ALL_ERRORS (1)

if `$result = false`
than `$result && $validator->isValid()` doesnt actually execute this function.  so i put this on a seperate line to execute first.

### DIFF
--- a/EmailValidator/Validation/MultipleValidationWithAnd.php
+++ b/EmailValidator/Validation/MultipleValidationWithAnd.php
@@ -62,7 +62,8 @@ class MultipleValidationWithAnd implements EmailValidation
         $errors = [];
         foreach ($this->validations as $validation) {
             $emailLexer->reset();
-            $result = $result && $validation->isValid($email, $emailLexer);
+            $validationResult = $validation->isValid($email, $emailLexer);
+            $result = $result && $validationResult;
             $this->warnings = array_merge($this->warnings, $validation->getWarnings());
             $errors = $this->addNewError($validation->getError(), $errors);
 
--- a/Tests/EmailValidator/Validation/MultipleValidationWithAndTest.php
+++ b/Tests/EmailValidator/Validation/MultipleValidationWithAndTest.php
@@ -10,7 +10,7 @@ use Egulias\EmailValidator\Warning\AddressLiteral;
 use Egulias\EmailValidator\Warning\DomainLiteral;
 use PHPUnit\Framework\TestCase;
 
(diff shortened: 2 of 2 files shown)


EmailValidator/Validation/MultipleValidationWithAnd.php
```diff
@@ -62,7 +62,8 @@ public function isValid($email, EmailLexer $emailLexer)
         $errors = [];
         foreach ($this->validations as $validation) {
             $emailLexer->reset();
-            $result = $result && $validation->isValid($email, $emailLexer);
+            $validationResult = $validation->isValid($email, $emailLexer);
+            $result = $result && $validationResult;
```
peter279k: How about using `$result = ($result && $validationResult);` to make the code readability?.

egulias:
Hi @joelharkes , I see your point, nice catch! Would you mind improving the PR with:

- [ ] Add a test case to prove the error, solved by your fix. If you look at the [current test](https://github.com/egulias/EmailValidator/blob/master/Tests/EmailValidator/Validation/MultipleValidationWitAndTest.php#L74)  it is only testing a valid email. The bug only happens when `$result = flase` and `self::ALLOW_ALL_ERRORS` for an invalid email.
- [ ] Fix the test class typo :) (ti says ...WitAnd when it should be Wit**h**And )
- [ ] Check the behaviour for `$result = flase` and `self::STOP_ON_ERROR` for an invalid email, is it adding one more error than expected?

EmailValidator/Validation/MultipleValidationWithAnd.php
joelharkes: if that's general code style i can change it, but for me doesn't make it more readable.

joelharkes:
checked firs boxes, not sure about third thing you mentioned what you mean exactly.

joelharkes:
Had some issues wrapping my head around the unit testing framework, experience in php is low, got it to work => fixed the test it by expecting to call once() instead of any().

egulias:
Hi @joelharkes , nice work! Sorry for my delay.
About the third box, is a doubt I have:
How does this code does when  `$result = false` and `self::STOP_ON_ERROR` for an invalid email?
Will it add one more error than expected?
Do you mind adding a test to check this does not happens?

egulias:
Thankyou very much @joelharkes !

Issue #169, opened 2018-04-11
[Insight] Unused use statement should be avoided - in …/MultipleValidationWithAnd.php, line 6


Saibamen:
**[in EmailValidator/Validation/MultipleValidationWithAnd.php, line 6](https://github.com/egulias/EmailValidator/blob/19811e00829958f06d9bfbfeb0d0e6b7aaff3767/EmailValidator/Validation/MultipleValidationWithAnd.php#L6)**

> The class `Egulias\EmailValidator\Exception\InvalidEmail` is declared but never used. You should remove the `use` statement.

```php
<?php

namespace Egulias\EmailValidator\Validation;

use Egulias\EmailValidator\EmailLexer;
use Egulias\EmailValidator\Exception\InvalidEmail;
use Egulias\EmailValidator\Validation\Exception\EmptyValidationList;

class MultipleValidationWithAnd implements EmailValidation
{
    /**
```

---------------------

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/22ba6692-9c02-42e5-a65d-1c5696bfffc6/analyses/31)_

Developerayo:
#Fixed https://github.com/Developerayo/EmailValidator

egulias:
Please make a PR then!
Thanks!

GrahamCampbell:
This can be closed now?

egulias:
Solved.

Issue #168, opened 2018-04-11
[Insight] Unused use statement should be avoided - in EmailValidator/Parser/LocalPart.php, line 8


Saibamen:
**[in EmailValidator/Parser/LocalPart.php, line 8](https://github.com/egulias/EmailValidator/blob/19811e00829958f06d9bfbfeb0d0e6b7aaff3767/EmailValidator/Parser/LocalPart.php#L8)**

> The class `Egulias\EmailValidator\EmailValidator` is declared but never used. You should remove the `use` statement.

```php
namespace Egulias\EmailValidator\Parser;

use Egulias\EmailValidator\Exception\DotAtEnd;
use Egulias\EmailValidator\Exception\DotAtStart;
use Egulias\EmailValidator\EmailLexer;
use Egulias\EmailValidator\EmailValidator;
use Egulias\EmailValidator\Exception\ExpectingAT;
use Egulias\EmailValidator\Exception\ExpectingATEXT;
use Egulias\EmailValidator\Exception\UnclosedQuotedString;
use Egulias\EmailValidator\Exception\UnopenedComment;
use Egulias\EmailValidator\Warning\CFWSWithFWS;
```

---------------------

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/22ba6692-9c02-42e5-a65d-1c5696bfffc6/analyses/31)_

Issue #125, opened 2016-07-29
Use EmailLexer::class sintax for retrieveng class FQCN in Tests


egulias:
E.g https://github.com/egulias/EmailValidator/blob/master/Tests/EmailValidator/Validation/MultipleValidationWitAndTest.php#L16

dereuromark:
:+1:

PR #190, opened 2019-01-19
Enabling psalm on php >= 7.1, fixing #125, #168, #169, partially fixing #177

### LINKED ISSUE #177: Using polyfill for ext-intl if

SignpostMarv:
* re: 210ac3c & da2e0c6, there's an `@psalm-suppress` or two in there rather than a baseline file because the older versions of psalm required for `php < 7` consider the baseline attribute invalid so that'd be best put off until you make the decision to drop support for php 5 in a given release.
* c504af2 only partially resolves #177 because the `SpoofChecker` class isn't polyfilled. The options there would be to find a polyfill just for SpoofChecker, or to fork out the `SpoofCheckValidation` class into a package that explicitly requires `ext-intl:*`
* 6812a28 & 02a6193 were done for value-added giggles.

### DIFF
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -15,10 +15,10 @@ class DNSCheckValidation implements EmailValidation
     private $warnings = [];
 
     /**
-     * @var InvalidEmail
+     * @var InvalidEmail|null
      */
     private $error;
-    
+
     public function __construct()
     {
         if (!extension_loaded('intl')) {
@@ -49,6 +49,11 @@ class DNSCheckValidation implements EmailValidation
         return $this->warnings;
     }
 
+    /**
+     * @param string $host
+     *
+     * @return bool
+     */
     protected function checkDNS($host)
     {
         $variant = INTL_IDNA_VARIANT_2003;
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -21,7 +21,7 @@ class DNSCheckValidation implements EmailValidation
 
     public function __construct()
     {
-        if (!extension_loaded('intl')) {
+        if (!function_exists('idn_to_ascii')) {
             throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
   },
   "require": {
     "php": ">=5.5",
-    "doctrine/lexer": "^1.0.1"
+    "doctrine/lexer": "^1.0.1",
+    "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
     "satooshi/php-coveralls": "^1.0.1",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
     "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
-    "ext-intl": "*",
     "satooshi/php-coveralls": "^1.0.1",
     "phpunit/phpunit": "^4.8.36|^7.5.15",
     "dominicsayers/isemail": "^3.0.7"
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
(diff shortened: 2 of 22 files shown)


egulias:
Hi @SignpostMarv great PR, thanks. I'll take a deeper look into Psalm. However, I'm relying on [Scrutinizer](https://scrutinizer-ci.com/g/egulias/EmailValidator/) for static analysis and would like to know if it is worth to have both.

SignpostMarv:
I've been using a combination of phpstan, psalm, and scrutinizer for quite a while now- my experience is that scrutinizer doesn't catch everything and that running static analysis on your local machine leads to a more rapid development cycle as you're not waiting for CI to spin up multiple instances & finish all the matrices. This leaves scrutinizer's main purpose down to tracking overall code quality & code coverage.

sebastian/phpcpd can be useful in identifying deduplication issues before they reach scrutinizer.

povils/phpmnd can help identify integers that should probably be consts.

maglnet/composer-require-checker checks if your code uses any php extensions or downstream code not explicitly required by the project.

k00ni:
When do you plan to merge this?

EmailValidator/Validation/DNSCheckValidation.php
```diff
@@ -15,13 +15,13 @@ class DNSCheckValidation implements EmailValidation
...
+     * @var InvalidEmail|null
      */
     private $error;
-    
+
     public function __construct()
     {
-        if (!extension_loaded('intl')) {
+        if (!function_exists('idn_to_ascii')) {
             throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
```
egulias: How does `idn_to_ascii` correlates to `intl` extension? Are they the same?
If so, please also update the error message.

composer.json
```diff
@@ -13,19 +13,15 @@
       "dev-master": "2.1.x-dev"
     }
   },
-  "repositories": [
-    {
-      "type": "git",
-      "url": "https://github.com/dominicsayers/isemail"
```
egulias: Why are you removing this? it is needed for the tests comparision :)

EmailValidator/Validation/DNSCheckValidation.php
SignpostMarv: `symfony/polyfill-intl-idn` will take over when the extension isn't loaded, thus the condition needs to check for the function used.

composer.json
SignpostMarv: as mentioned in the commit message, the package is listed on packagist & directly listing the repo causes install issues on windows & git bash.

GrahamCampbell:
This PR does a lot of different things, which makes it hard to merge. I think it should be broken up into more atomic pieces. Not all of the changes are still relevant now too.

SignpostMarv:
checking a rebase builds before splitting things off.

egulias:
@SignpostMarv apparently Psalm version you are requiring does not supports php 7.0 (https://travis-ci.org/egulias/EmailValidator/jobs/631128083?utm_medium=notification&utm_source=github_status)

SignpostMarv:
> @SignpostMarv apparently Psalm version you are requiring does not supports php 7.0 (https://travis-ci.org/egulias/EmailValidator/jobs/631128083?utm_medium=notification&utm_source=github_status)

no version constraint is applied, so this may occasionally fail (necessitating an adjustment to the matrix as in 079ac0e)

composer.json
```diff
@@ -14,9 +14,11 @@
   },
   "require": {
     "php": ">=5.5",
-    "doctrine/lexer": "^1.0.1"
+    "doctrine/lexer": "^1.0.1",
+    "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
+    "ext-intl": "*",
```
egulias: Relates to #231 , since intl is still required for the spoof checker.
Maybe you can revert this change to avoid blocking this PR because of the INTL dependency fix.

SignpostMarv: re: `SpoofChecker`, what're your thoughts [regarding forking `SpoofCheckValidation` to it's own package that requires `ext-intl`](https://github.com/egulias/EmailValidator/pull/229#discussion_r362545036) ?

egulias: Answered. However, we can still remove this changes from the current PR in order to merge it and solve the INTL issue there.

SignpostMarv: removed in affd3dd

PR #193, opened 2019-04-05
Include route-addr support for RFC 822 and 1123

jrchamp:
Test valid addresses that use:
 - the mailbox format from RFC 822: phrase route-addr
 - the mailbox format from RFC 1123 (phrase optional): [phrase] route-addr

Hopefully the automated tests will show whether we're correctly supporting the RFC addresses in the mailbox route-addr format.

### DIFF
--- a/Tests/EmailValidator/Validation/RFCValidationTest.php
+++ b/Tests/EmailValidator/Validation/RFCValidationTest.php
@@ -71,6 +71,8 @@ class RFCValidationTest extends TestCase
     public function getValidEmails()
     {
         return array(
+            ['"Fabien Potencier" <fabien@symfony.com>'],
+            ['<fabien@symfony.com>'],
             ['â@iana.org'],
             ['fabien@symfony.com'],
             ['example@example.co.uk'],


egulias:
Hi @jrchamp 
it looks like we are not, as stated in the readme: supported RFCs are 5321, 5322, 6530, 6531, 6532.

Are you suggesting/asking to add support for this format?

jrchamp:
Sorry, I did not see that RFC 822 had been obsoleted by RFC 2822, which has been further obsoleted by RFC 5322, which you do support. From RFC 5322:

- `address` is either a `mailbox` or a `group`
- `mailbox` is either `addr-spec` or `name-addr`
- `name-addr` is an optional `display-name` followed by an `angle-addr`
- `angle-addr` is roughly the equivalent of the the older `route-addr`, surrounding the `addr-spec` in angle brackets

If the goal of the EmailValidator software is the validate `address` or `mailbox` as specified in RFC 5322, then yes, I am asking for support for the `name-addr` syntax in addition to the existing `addr-spec` syntax. If the goal is instead to only validate the `addr-spec` syntax, then it would be helpful to clarify that in the documentation.

Thank you for your quick and helpful response!

egulias:
Hi @jrchamp , it took me some more this time.
Added your request as issue #195 
Thanks for the clear documentation.

PR #194, opened 2019-04-16
Add mailbox validator

IvanLutokhin:


EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    protected function getMXHosts($email)
+    {
+        $variant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : INTL_IDNA_VARIANT_2003;
+
+        $hostname = $email;
+        if ( false !== ($lastAtPos = strrpos($email, '@')) ) {
+            $hostname = substr($email, $lastAtPos + 1);
+        }
+        $hostname = rtrim(idn_to_ascii($hostname, IDNA_DEFAULT, $variant), '.') . '.';
```
egulias: Please avoid mutating variables this way, it is hard to follow.
Maybe
```php
$hostname = $this->extractHostname($email);

private function extractHostname($email)
{
  $lastAtPos = (bool) strrpos($email, '@')
  $variant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : INTL_IDNA_VARIANT_2003;
  if ($lastAtPos) { 
    $removedAT = substr($email, $lastAtPos + 1);
    return rtrim(idn_to_ascii($removedAT, IDNA_DEFAULT, $variant), '.') . '.';
  }
  return rtrim(idn_to_ascii($email, IDNA_DEFAULT, $variant), '.') . '.';
}
```

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+
+        $mxHosts = [];
+        $result = getmxrr($hostname, $mxHosts);
+        if ( ! $result ) {
```
egulias: Please remove space between `!` and `$` in all the files :).

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * @param string $hostname
+     * @param int $port
+     * @param int $timeout
+     * @param string $fromEmail
+     * @param string $toEmail
+     * @return bool
+     */
+    protected function checkMailbox($hostname, $port, $timeout, $fromEmail, $toEmail)
```
egulias: This params are already provided by properties of the class, except for `$hostname` and `$toEmail`.
Also, please make the method private.

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+    private function assertResponse($socket, $expectedCode)
+    {
+        if ( ! is_resource($socket) ) {
+            return false;
+        }
+
+        $data = '';
+        while (substr($data, 3, 1) !== ' ') {
+            if ( ! ( $data = @fgets($socket, 256) ) ) {
+                $this->lastResponseCode = -1;
```
egulias: Do you actually need this to be class level property? I haven't found this to be "reused", just overwritten. Making it local to the scope will prevent errors or unexpected manipulations from the outside.

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
```diff
@@ -0,0 +1,40 @@
...
+        $validation = new MailboxCheckValidation();
+        $expectedWarnings = [NoDNSMXRecord::CODE => new NoDNSMXRecord()];
+        $validation->isValid('example@invalid.example.com', new EmailLexer());
+        $this->assertEquals($expectedWarnings, $validation->getWarnings());
+    }
+
+    public function testIllegalMailboxError()
+    {
+        $validation = new MailboxCheckValidation();
+        $expectedError = new IllegalMailbox(550);
```
egulias: Why where you expecting code `550`? What does it means. This test is failing.

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
```diff
@@ -0,0 +1,40 @@
...
+use Egulias\EmailValidator\Validation\MailboxCheckValidation;
+use Egulias\EmailValidator\Warning\NoDNSMXRecord;
+use PHPUnit\Framework\TestCase;
+
+class MailboxCheckValidationTest extends TestCase
+{
+    public function testValidMailbox()
+    {
+        $validation = new MailboxCheckValidation();
+        $this->assertTrue($validation->isValid('no-reply@gmail.com', new EmailLexer()));
```
egulias: This test is also failing and I'd argue is quite fragile, as it depends on an external resource. Also not sure you can reach gmail on port 25 :).
You could use/implement one for the test (from a quick search, you could use https://github.com/axllent/fake-smtp).

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     * @inheritDoc
+     */
+    public function isValid($email, EmailLexer $emailLexer)
+    {
+        $mxHosts = $this->getMXHosts($email);
+
+        $isValid = false;
+        foreach ($mxHosts as $mxHost) {
+            if ( ($isValid = $this->checkMailbox($mxHost, $this->port, $this->timeout, $this->fromEmail, $email)) ) {
+                break;
```
egulias: For the sake of clarity, please change to
```php
if($this->checkMailbox(...)) {
  $isValid = true;
  break;
}
```

EmailValidator/Validation/MailboxCheckValidation.php
egulias: Maybe a better name, given the bool returned, would be `mailboxExist`.

egulias (changes requested):
Hi @IvanLutokhin ,
really appreciate your PR. I did an exhaustive review, please look at the comments. Let me know any doubt or further comment on the review.
Great work!

EmailValidator/Validation/Error/IllegalMailbox.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * IllegalMailbox constructor.
+     *
+     * @param int $responseCode
+     */
+    public function __construct($responseCode)
+    {
+        parent::__construct();
+
+        $this->responseCode = $responseCode;
```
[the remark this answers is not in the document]
IvanLutokhin: In this place private var is necessary because `responseCode` contains SMTP response code https://www.greenend.org.uk/rjk/tech/smtpreplies.html

Users can understand the reason, why mailbox is illegal.

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public function getWarnings()
+    {
+        return $this->warnings;
+    }
+
+    /**
+     * @return int
+     */
+    public function getLastResponseCode()
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public function getPort()
+    {
+        return $this->port;
+    }
+
+    /**
+     * @param int $port
+     */
+    public function setPort($port)
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public function getTimeout()
+    {
+        return $this->timeout;
+    }
+
+    /**
+     * @param int $timeout
+     */
+    public function setTimeout($timeout)
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public function getFromEmail()
+    {
+        return $this->fromEmail;
+    }
+
+    /**
+     * @param string $fromEmail
+     */
+    public function setFromEmail($fromEmail)
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * @var int
+     */
+    private $timeout = 10;
+
+    /**
+     * @var string
+     */
+    private $fromEmail = 'test-mailbox@validation.email';
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
```diff
@@ -0,0 +1,240 @@
...
+     * @param string $email
+     *
+     * @return array
+     */
+    protected function getMXHosts($email)
+    {
+        $variant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : INTL_IDNA_VARIANT_2003;
+
+        $hostname = $email;
+        if ( false !== ($lastAtPos = strrpos($email, '@')) ) {
```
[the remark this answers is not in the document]
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
IvanLutokhin: fix

EmailValidator/Validation/MailboxCheckValidation.php
IvanLutokhin: Yes, is needed property.

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
IvanLutokhin: 550 - it is smtp response code
`mailbox unavailable`

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
IvanLutokhin: fix tests for use mock

EmailValidator/Validation/MailboxCheckValidation.php
IvanLutokhin: fix

IvanLutokhin:
Thanks for exhaustive review.
I try to fix all issue and comments it.

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
egulias: Thanks for the clarification. Do you mind leaving a comment in the code please?

EmailValidator/Validation/MailboxCheckValidation.php
egulias: I think my comment was misleading. The variable is needed, that´s correct. However, instead of using a class property I think it can be a local (that is, declared and used within the method) var.
Using
```php
function Bar {
  $lastResponseCode = null;
}
```
Instead of 
```php
class Foo {
  private $lastResponseCode;
  function Baz {
     $this->lastResponseCode...
  }
}
```

Tests/EmailValidator/Validation/MailboxCheckValidationTest.php
egulias: Thanks for the clarification. Do you mind leaving a comment in the code please?

EmailValidator/Validation/Error/IllegalMailbox.php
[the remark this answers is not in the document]
egulias: You are right.
What happens is that `InvalidEmail` does not provide a method for accessing `responseCode`, which is private. Thus, user won't be able to get the code.
The only way for user, AFAIK, to get the code would be to overwrite intherited `__toString` (see https://www.php.net/manual/en/class.exception.php and https://www.php.net/manual/en/language.exceptions.extending.php), the only method allowed to be overwritten:
```php
public function __toString()
{
  return ": SMTP response code was {$this->responseCode}, exception reason,code {$this->message} - {$this->code}";
}
```
Another solution would be create an empty interface `EmailValidationException`, make `InvalidEmail` implement it and then 
```php
InvalidMailbox extends \InvalidArgumentException implements EmailValidationException
{
    private $reasons = [550 => 'Requested action not taken: mailbox unavailable', 551 => ...]
    public function __construct($responseCode)
    {
        $selectedReason = isset($reasons[$responseCode]) ? $reasons[$responseCode] : "Unknown reason";
        parent::__construct($selectedReason, $responseCode);
    }
}
```
And adapt all the necesary code to allow for this new type of exception by requiring the interface `EmailValidationException` instead of the concrete class `InvalidEmail`.
```
What do you think?

egulias (changes requested):

egulias:
Thanks for the changes @IvanLutokhin . I have done a couple of comments more. Please also check the test failure for PHP 5.5.

IvanLutokhin:
Yes, I added getter for property _responseCode_ in _InvalidMailbox_ and resolve magic method ___toString()_
I think that's enough and interface is not necessary.

I fixed unit test for php 5.5 also. The problem was in deprecated methods phpunit.

egulias:
Apparently there´s an error for hhvm within Travis, hence the failure. But the rest is passing.

egulias:
@IvanLutokhin regarding the exception, ___toString_ is a potential solution. However this would be the only error exception that has a _getResponseCode_ method so if someone wants to do something with it the only way would be to test against the class name of the returned error. By providing an specific type that would be easier.
It will also allow you to provide a wider range of error responses and codes following the table you shared, which would be great to have.
Happy to find a solution together!
Thanks again for the time you are investing into the feature!

IvanLutokhin:
> Apparently there´s an error for hhvm within Travis, hence the failure. But the rest is passing.

HHVM does not support composer anymore.
https://hhvm.com/blog/2019/02/11/hhvm-4.0.0.html

As one of the solutions to fix the version HHVM 3.30.

egulias:
Thanks @IvanLutokhin , I'll do that. I'll be working to merge the PR shortly. Thanks for your patience.

osavchenko:
@IvanLutokhin do you need any help on this PR?

clemlatz:
@egulias @IvanLutokhin I'm interested in the feature.
Any reasons why this PR hasn't been merged? Anything I can do to help?

egulias:
Hi folks, @IvanLutokhin , apologies for the delay here. I have fixed the checks so please update with the latest code to have the checks pass.
Happy to look at this again. If anyone is still interested either on this PR or on a new one forking this one.

Cheers.

egulias:
BTW `master` is no longer a valid branch. Please PR to latest version.

PR #198, opened 2019-05-06
Simplify installation command in README

bocharsky-bw:
Composer will automatically use BC constraint, like `^2.1.7`, so I think specifying version explicitly is redundant


egulias:
As said in another PR, Travis is failing internally on HHVM, not because of tests.

egulias:
Thanks @bocharsky-bw

PR #199, opened 2019-05-14
PHP Bug: https://bugs.php.net/bug.php?id=77827

amenophis:
The test suite doesn't run well on php 7.3.5, but travis run the 7.3.2.
There is a change since php 7.3.5 (https://bugs.php.net/bug.php?id=77827).
what can be done to fix that ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+Run CI
+
 # EmailValidator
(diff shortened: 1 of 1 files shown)


egulias:
Hi @amenophis , thanks for the issue request. First of all the PR should include the change in travis.yaml to trigger a 7.3.5 failure so we can pin the PR to an issue so the next person comming can see the problem.
I have read the bug report and it is solved, so I expect in version 7.3.6 to be fixed. Thus we can recommend using 7.3.6.

So, do you mind:
1. Do a PR triggering the right CI build failure
2. Open an issue with a descriptive title, something like "PHP 7.3.5 bug in preg_match" and link to this PR.

Thanks!

egulias:
Thanks @amenophis , looks like the bug has been solved and CI no longer fails. Thanks!

PR #203, opened 2019-07-18
Fix PHP 7.4 support

nicolas-grekas:
Symfony is going to need this to have green CI on PHP 7.4 :)

### DIFF
--- a/EmailValidator/EmailLexer.php
+++ b/EmailValidator/EmailLexer.php
@@ -77,10 +77,22 @@ class EmailLexer extends AbstractLexer
 
     protected $previous;
 
+    private static $nullToken = [
+        'value' => '',
+        'type' => null,
+        'position' => 0,
+    ];
+
+    public function __construct()
+    {
+        $this->previous = $this->token = self::$nullToken;
+    }
+
     public function reset()
     {
         $this->hasInvalidTokens = false;
         parent::reset();
+        $this->previous = $this->token = self::$nullToken;
     }
 
     public function hasInvalidTokens()
@@ -122,8 +134,10 @@ class EmailLexer extends AbstractLexer
     public function moveNext()
     {
         $this->previous = $this->token;
+        $hasNext = parent::moveNext();
+        $this->token = $this->token ?: self::$nullToken;
 
-        return parent::moveNext();
+        return $hasNext;
     }
 
     /**
--- a/Tests/EmailValidator/Validation/MultipleValidationWithAndTest.php
+++ b/Tests/EmailValidator/Validation/MultipleValidationWithAndTest.php
@@ -2,6 +2,7 @@
 
 namespace Egulias\Tests\EmailValidator\Validation;
 
+use Egulias\EmailValidator\EmailLexer;
 use Egulias\EmailValidator\EmailValidator;
 use Egulias\EmailValidator\Exception\CommaInDomain;
 use Egulias\EmailValidator\Exception\NoDomainPart;
@@ -17,7 +18,7 @@ class MultipleValidationWithAndTest extends TestCase
 {
     public function testUsesAndLogicalOperation()
     {
-        $lexer = $this->getMockBuilder("Egulias\\EmailValidator\\EmailLexer")->getMock();
+        $lexer = new EmailLexer();
         $validationTrue = $this->getMockBuilder("Egulias\\EmailValidator\\Validation\\EmailValidation")->getMock();
         $validationTrue->expects($this->any())->method("isValid")->willReturn(true);
         $validationTrue->expects($this->any())->method("getWarnings")->willReturn([]);
@@ -38,7 +39,7 @@ class MultipleValidationWithAndTest extends TestCase
 
     public function testValidationIsValid()
     {
-        $lexer = $this->getMockBuilder("Egulias\\EmailValidator\\EmailLexer")->getMock();
+        $lexer = new EmailLexer();
 
         $validation = $this->getMockBuilder("Egulias\\EmailValidator\\Validation\\EmailValidation")->getMock();
         $validation->expects($this->any())->method("isValid")->willReturn(true);
@@ -59,7 +60,7 @@ class MultipleValidationWithAndTest extends TestCase
         ];
         $expectedResult = array_merge($warnings1, $warnings2);
 
-        $lexer = $this->getMockBuilder("Egulias\\EmailValidator\\EmailLexer")->getMock();
+        $lexer = new EmailLexer();
         $validation1 = $this->getMockBuilder("Egulias\\EmailValidator\\Validation\\EmailValidation")->getMock();
         $validation1->expects($this->any())->method("isValid")->willReturn(true);
         $validation1->expects($this->once())->method("getWarnings")->willReturn($warnings1);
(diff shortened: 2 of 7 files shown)


nicolas-grekas:
Ready.

EmailValidator/EmailLexer.php
```diff
@@ -77,10 +77,16 @@ class EmailLexer extends AbstractLexer
 
     protected $previous;
 
+    public function __construct()
+    {
+        $this->reset();
```
egulias: Calling `reset` on construction sounds a bit odd, reseting something 'new'. For what I've seen in the rest of the changes, this is needed mainly to do the equality evaluation (e.g https://github.com/egulias/EmailValidator/pull/203/files#diff-c759db088d5c729c1f537d85c2ea93d0R187).
I would suggest something like `init()` :
```php
$this->hasInvalidTokens = false;
$this->previous = $this->token = ['type' => null, 'value' => '', 'position' => 0];
```
and call that from both `__contruct` & `reset`.

EmailValidator/EmailLexer.php
```diff
@@ -77,10 +77,16 @@ class EmailLexer extends AbstractLexer
...
+    }
+
     public function reset()
     {
         $this->hasInvalidTokens = false;
         parent::reset();
+        $this->previous = $this->token = ['type' => null, 'value' => '', 'position' => 0];
```
egulias: I'd suggest to use `EmailLexer::C_NUL` for `type` instead.

Tests/EmailValidator/Validation/MultipleValidationWithAndTest.php
```diff
@@ -59,7 +60,7 @@ public function testAccumulatesWarnings()
         ];
         $expectedResult = array_merge($warnings1, $warnings2);
 
-        $lexer = $this->getMockBuilder("Egulias\\EmailValidator\\EmailLexer")->getMock();
+        $lexer = new EmailLexer();
```
egulias: 👍

EmailValidator/EmailLexer.php
```diff
@@ -122,8 +128,10 @@ public function getPrevious()
     public function moveNext()
     {
         $this->previous = $this->token;
+        $hasNext = parent::moveNext();
+        $this->token = $this->token ?: ['type' => null, 'value' => '', 'position' => 0];
```
egulias: Is this line because of the lexer reaching the last token and being set to `null`?
I'd suggest to express that case with something like 
```php
public function moveNext()
{
    $this->previous = $this->token;
    $hasNext = parent::moveNext();
    /* new stuff */ $this->setEndToken($hasNext);
    return $hasNext;
}

private function setFinalToken($nextToken)
{
    $this->token = $this->token ?: ['type' => C_NUL, 'value' => '', 'position' => 0];
}
```

phpunit.xml.dist
```diff
@@ -19,8 +18,12 @@
...
-  <blacklist>
-    <directory>./vendor</directory>
-  </blacklist>
+  <whitelist>
+    <directory>./EmailValidator/</directory>
+  </whitelist>
 </filter>
+
+<listeners>
+  <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
```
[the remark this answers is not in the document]
nicolas-grekas: because PHP 7.4 introduces new deprecation that affect phpunit, and the listener allows silencing them

EmailValidator/EmailLexer.php
nicolas-grekas: the code is full of checks with `$this->lexer->token['type']`, when `$this->lexer->token` is `null` actually. PHP 7.4 is now refusing to access keys on nul values, so that the affected properties must be set to an array representing the null-state. reset does it.

EmailValidator/EmailLexer.php
nicolas-grekas: this wouldn't be the same - we don't want to this be confused with the NUL char.

EmailValidator/EmailLexer.php
nicolas-grekas: you're right about the reason: $this->token cannot be null.
not sure I get what the additional method would provide thought :)

EmailValidator/EmailLexer.php
egulias: Thanks for the explanation, understood. My suggestion is moving both lines to a new method so they can call it having a more expresive meaning and avoiding coupling too construction with re-use (reseting to analyse another string).

```php
__construct()
{
    $this->init();
}

reset()
{
...
$this->init();
}
```

EmailValidator/EmailLexer.php
egulias: Agree. Should a new token be added?

EmailValidator/EmailLexer.php
egulias: To understood this line I had to review carefully the full PR and take my time, even with my knowledge of the library.
The additional method adds clarity on the reason behind.

nicolas-grekas: I pushed a variant, does it solve the concern?

EmailValidator/EmailLexer.php
nicolas-grekas: I don't think so, `null` is really the best "const" for this to me.

EmailValidator/EmailLexer.php
egulias: I think it can work.
Thanks!

SamMousa:
Is this really the way to go about this @nicolas-grekas ??
It seems weird to me for 3rd party libraries to be required to include some dev package from symfony just so Symfony can have their tests pass...

nicolas-grekas:
@SamMousa you ist be confused, this added dev dep is not to make Symfony's tests pass, but about making this very package tests pass.

SamMousa:
I have just removed it locally, the test listener and the dependency, and tests still pass.

nicolas-grekas:
Try again with PHP7.4 + latest phpunit.

SamMousa:
```bash
root@ba1517249f5f:/project# php --version
PHP 7.4.0alpha3 (cli) (built: Jul 12 2019 21:26:03) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0-dev, Copyright (c) Zend Technologies

PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

...............................................................  63 / 253 ( 24%)
............................................................... 126 / 253 ( 49%)
............................................................... 189 / 253 ( 74%)
...........................................................SSSS 252 / 253 ( 99%)
.                                                               253 / 253 (100%)

Time: 537 ms, Memory: 6.00MB

OK, but incomplete, skipped, or risky tests!
Tests: 253, Assertions: 351, Skipped: 4.
```

SamMousa:
Works for me:
```bash
root@ba1517249f5f:/project# php --version
PHP 7.4.0alpha3 (cli) (built: Jul 12 2019 21:26:03) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0-dev, Copyright (c) Zend Technologies

PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

...............................................................  63 / 253 (
24%)
............................................................... 126 / 253 (
49%)
............................................................... 189 / 253 (
74%)
...........................................................SSSS 252 / 253 (
99%)
.                                                               253 / 253
(100%)

Time: 537 ms, Memory: 6.00MB

OK, but incomplete, skipped, or risky tests!
Tests: 253, Assertions: 351, Skipped: 4.
```

nicolas-grekas:
Oh that's good news, maybe related to the removed mocks? I suppose I
patched then after adding the bridge and didn't try removing it after. PR
welcome I suppose.

mlocati:
@egulias would you accept a similar pull request for version 1?

egulias:
Hi @mlocati 
Yes, happy do it. It is still in use from what Packagist says.

mlocati:
Done: see #206

PR #225, opened 2019-12-30
[2.1] Update .gitignore

GrahamCampbell:


egulias:
Hi Graham, 
could removing the `.lock` file generate inconsitencies when running tests or using the lib?

GrahamCampbell:
Composer totally ignores the lock file when you install the library, and the tests should always work on the newest dependencies, so there's no point in having the lock file (also, the lock file is already deleted from the repo anyway).

PR #226, opened 2019-12-30
[2.1] Removed unneeded repositories section from composer.json

GrahamCampbell:


egulias:
Hi Graham, 
this repositorie is required for this
https://github.com/egulias/EmailValidator/blob/master/Tests/Performance/AgainstOldIsemail.php.

What would you suggest?

GrahamCampbell:
The repository is not "required", since it is published on packagist. This why the tests still pass on travis. :)

GrahamCampbell:
Do you see what I mean?

egulias:
Yes, you are right. However if I recall correctly, performance tests are not run as part of the build, since they are mostly "informative".
Thanks for you PRs!

GrahamCampbell:
Right. So, I suppose I should say, this is why the compose install didn't fail on travis.

GrahamCampbell:
Since the test itself didn't actually run, as you say. :)

PR #229, opened 2019-12-31
loading in symfony polyfill, partially resolves #177

### LINKED ISSUE #177: Using polyfill for ext-intl if

SignpostMarv:
* loads a polyfill instead of relying directly on extension
* requires intl extension for local dev.

regarding the change from `extension_loaded` to `function_exists`, [see conversation](https://github.com/egulias/EmailValidator/pull/190/commits/cb793ad34b3738b016b5726d5a765fd1b75138aa) from #190

### DIFF
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -21,7 +21,7 @@ class DNSCheckValidation implements EmailValidation
     
     public function __construct()
     {
-        if (!extension_loaded('intl')) {
+        if (!function_exists('idn_to_ascii')) {
             throw new \LogicException(sprintf('The %s class requires the Intl extension.', __CLASS__));
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
   ],
   "require": {
     "php": ">=5.5",
-    "doctrine/lexer": "^1.0.1"
+    "doctrine/lexer": "^1.0.1",
+    "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
     "satooshi/php-coveralls": "^1.0.1",
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
     "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
+    "ext-intl": "*",
     "satooshi/php-coveralls": "^1.0.1",
     "phpunit/phpunit": "^4.8.36|^7.5.15",
     "dominicsayers/isemail": "^3.0.7"


composer.json
```diff
@@ -20,9 +20,11 @@
   ],
   "require": {
     "php": ">=5.5",
-    "doctrine/lexer": "^1.0.1"
+    "doctrine/lexer": "^1.0.1",
+    "symfony/polyfill-intl-idn": "^1.10"
   },
   "require-dev": {
+    "ext-intl": "*",
```
GrahamCampbell: Is that needed? Do the tests fail with only the polyfill?

GrahamCampbell: If they do, we should fix the implementation so they pass, and then remove this line. Otherwise, still remove this line.

egulias: I understand that by adding polyfill something in the implementation should be changed to use that instead of intl.
For example in https://github.com/egulias/EmailValidator/blob/135444003f7032fa45e3a688737fed4522cb9312/EmailValidator/Validation/SpoofCheckValidation.php#L26

EmailValidator/Validation/DNSCheckValidation.php
```diff
@@ -21,7 +21,7 @@ class DNSCheckValidation implements EmailValidation
     
     public function __construct()
     {
-        if (!extension_loaded('intl')) {
+        if (!function_exists('idn_to_ascii')) {
```
egulias: By adding Symfony's polyfill, this shouldn't be needed any more. Since it is a required dep and that dep resolves the intl requirement. However, the Spoofchecker issue (see below) still required intl.

GrahamCampbell:
I've prepared https://github.com/egulias/EmailValidator/pull/231, which replaces this PR.

GrahamCampbell:
My replaced PR should totally fix the original issue too.

composer.json
SignpostMarv: SpoofChecker isn't present in the polyfill.

GrahamCampbell: Can we ask for this to be implemented in the polyfill. Or, could we polyfill it ourselves in this package?

SignpostMarv: I am insufficiently familiar with how spoofchecker works to advise regarding polyfilling, so I'd imagine the two feasible options are:
1) enquire with symfony re: fleshing out the polyfill
2) fork the spoof check validation out into it's own package that requires `ext-intl:*`, negating the need to polyfill it.

EmailValidator/Validation/DNSCheckValidation.php
SignpostMarv: rendered moot by #231

composer.json
egulias: I find the 2nd option more appealing as is more in control of the validator. Looks like a bit overkill, having a package only for that, however leveragin the full benefit of the polyfill is interesting.

PR #231, opened 2020-01-02
[WIP] Fix intl support

### LINKED ISSUE #177: Using polyfill for ext-intl if

GrahamCampbell:
TODO:
  - [x] Add intl polyfill dependency
  - [x] Remove checks on extension loaded
  - [ ] Find or write a polyfill for Spoofchecker

---

Closes #177. Closes https://github.com/egulias/EmailValidator/pull/229.

### DIFF
--- a/EmailValidator/Validation/DNSCheckValidation.php
+++ b/EmailValidator/Validation/DNSCheckValidation.php
@@ -19,13 +19,6 @@ class DNSCheckValidation implements EmailValidation
      */
     private $error;
     
-    public function __construct()
-    {
-        if (!extension_loaded('intl')) {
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
@ohartl why have you approved this?

ohartl:
> @ohartl why have you approved this?

As this would solve #177, i reviewed this for my pr for my specific use case and it works for that.

@GrahamCampbell Except UndefinedClass Spoofchecker stuff, but that seems to be fixable on symfony/polyfill project side right?

GrahamCampbell:
Symfony have no plans to implement the spoofing class (which is totally fair enough), and so I don't think this PR can progress.

egulias:
Thanks for updating @GrahamCampbell

PR #242, opened 2020-06-06
Move files in typical src and tests folder

simonschaufi:


simonschaufi:
Surprisingly psalm throws the same errors in master branch as well so I didn't break it!

egulias:
Hi @simonschaufi  I think it makes sense.
It would be amazing if you fix the psalm error too ;).

simonschaufi:
To be honest, I don't understand the error message from psalm.

simonschaufi:
@egulias would you agree with merging this one and I'll try my best to find a solution in another PR

egulias:
I understand the problem is here https://github.com/egulias/EmailValidator/blob/c8d2ff978dfae1fbb48dbe64c1dbfeca780ce95b/src/Parser/Parser.php#L25
The docblock says that it is an array of "Warning" type, however, the comparission is being done against an int from a constant. I think it needs a dockblock of type "array" instead. 
Makes sense?

simonschaufi:
yes, this fixes it! Thank you for your help!

egulias:
Great news! All green! There we go.
Thanks @simonschaufi

Issue #249, opened 2020-07-09
DNS Validation Issue


BenHarris:
I hit an issue today where the DNS validation wasn't working. It looks like there is a bug in alpine Linux (I was using the php:7.4-fpm-alpine docker image), which means checkdnsrr returns true even if the domain doesn't exist. See the bug reports below:

- https://bugs.php.net/bug.php?id=78008
- https://gitlab.alpinelinux.org/alpine/aports/-/issues/7392

Whilst I appreciate this is not a bug in this library, I wonder if it might be worth switching to dns_get_record in DNSCheckValidation instead, which is unaffected, allows us the same validation and was also somewhat faster in my tests.

dns_get_record returns an array or false, so the equivalent implementation would just be to do something like `$MXresult = !empty(dns_get_record($host, DNS_MX));`

### Invalid domain test
| Function | Result | Runtime |
|-|-|-|
| `checkdnsrr('notavaliddomain.c2.', 'MX')` | bool(false) | 0.00630 |
| `dns_get_record('notavaliddomain.d2.', DNS_MX)` | array(0){} | 0.00497 |

### Valid domain
| Function | Result | Runtime |
|-|-|-|
| `checkdnsrr('google.com.', 'MX')` | bool(true) | 0.00152 |
| `dns_get_record('google.com.', DNS_MX)` | <pre>array(5) {<br>  [0]=> array(6) {<br>    ["host"]=> string(10) "google.com"<br>    ["class"]=> string(2) "IN"<br>    ["ttl"]=> int(600)<br>    ["type"]=> string(2) "MX"<br>    ["pri"]=> int(40)<br>["target"]=> string(23) "alt3.aspmx.l.google.com"<br>  }<br>  ...<br>}</pre> | 0.00115 |

BenHarris:
Another benefit to this approach would be handling the "Null MX" case [RFC 7505](https://tools.ietf.org/html/rfc7505).

`checkdnsrr('example.com.', 'MX')` returns `bool(true)` as the MX record exists. As there is no target, it should be considered invalid and return false. `dns_get_record()` returns the DNS records, so we can simply validate that a) there is an MX record and b) it is a valid record.

egulias:
Hi @BenHarris , 
Thanks for informing about this issue.
I see the benefits you state and i'd be a great improvement for that validation.
I encourage you to do the change and send a PR ;), I will be more than happy to review and merge.

Looking forward to that PR!

egulias:
[...]

PR #250, opened 2020-07-20
Reimplement DNSCheckValidation

BenHarris:
Utilise get_dns_record and extend support for the detection of Null MX records (RFC7505) and reserved, mDNS and private namespaces (RFC2606 & RFC6762). References #249.

I have had to remove the testDnsWarnings test as I'm not sure of any specific domains which have one or more A/AAAA records, but no MX records, which are not "Null MX" records.

### DIFF
--- a/src/Exception/DomainAcceptsNoMail.php
+++ b/src/Exception/DomainAcceptsNoMail.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Egulias\EmailValidator\Exception;
+
+class DomainAcceptsNoMail extends InvalidEmail
+{
+    const CODE = 154;
+    const REASON = 'Domain accepts no mail (Null MX, RFC7505)';
+}
--- a/src/Exception/LocalOrReservedDomain.php
+++ b/src/Exception/LocalOrReservedDomain.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Egulias\EmailValidator\Exception;
+
+class LocalOrReservedDomain extends InvalidEmail
+{
+    const CODE = 153;
+    const REASON = 'Local, mDNS or reserved domain (RFC2606, RFC6762)';
+}
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -21,6 +21,12 @@ class DNSCheckValidation implements EmailValidation
      */
     private $error;
 
+    /**
+     * @var array
+     */
+    private $mxRecords = [];
+
+
     public function __construct()
     {
         if (!function_exists('idn_to_ascii')) {
@@ -68,7 +74,7 @@ class DNSCheckValidation implements EmailValidation
             return false;
         }
 
-        return $this->checkDNS($host);
+        return $this->checkDns($host);
     }
 
     public function getError()
@@ -86,48 +92,71 @@ class DNSCheckValidation implements EmailValidation
      *
      * @return bool
      */
-    protected function checkDNS($host)
+    protected function checkDns($host)
     {
         $variant = INTL_IDNA_VARIANT_2003;
         if (defined('INTL_IDNA_VARIANT_UTS46')) {
             $variant = INTL_IDNA_VARIANT_UTS46;
         }
+
         $host = rtrim(idn_to_ascii($host, IDNA_DEFAULT, $variant), '.') . '.';
 
-        // Get all MX, A and AAAA DNS records
+        return $this->validateDnsRecords($host);
+    }
+
+
+    /**
+     * Validate the DNS records
+     *
+     * @param $dnsRecords A set of DNS records in the format returned by dns_get_record.
+     *
+     * @return bool True on success.
+     */
+    private function validateDnsRecords($host)
+    {
+        // Get all MX, A and AAAA DNS records for host
         $dnsRecords = dns_get_record($host, DNS_MX + DNS_A + DNS_AAAA);
 
-        // No MX, A or AAAA records
+
+        // No MX, A or AAAA DNS records
         if (empty($dnsRecords)) {
             $this->error = new NoDNSRecord();
             return false;
         }
 
-        $aRecords  = [];
-        $mxRecords = [];
-
-        // Iterate over all returned DNS records
+        // For each DNS record
         foreach ($dnsRecords as $dnsRecord) {
-            if ($dnsRecord['type'] == 'MX') {
-                // "Null MX" record indicates the domain accepts no mail (https://tools.ietf.org/html/rfc7505)
-                if (empty($dnsRecord['target']) || $dnsRecord['target'] == '.') {
-                    $this->error = new DomainAcceptsNoMail();
-                    return false;
-                }
-
-                $mxRecords[] = $dnsRecord;
-                continue;
-            }
-
-            if (in_array($dnsRecord['type'], ['A', 'AAAA'])) {
-                $aRecords[] = $dnsRecord;
+            if (!$this->validateMXRecord($dnsRecord)) {
+                return false;
             }
         }
 
-        // No MX record
-        if (empty($mxRecords)) {
+        // No MX records (fallback to A or AAAA records)
+        if (empty($this->mxRecords)) {
             $this->warnings[NoDNSMXRecord::CODE] = new NoDNSMXRecord();
         }
+
+        return true;
+    }
+
+    /**
+     * Validate an MX record
+     *
+     * @param $dnsRecord A DNS record.
+     */
+    private function validateMxRecord($dnsRecord)
+    {
+        if ($dnsRecord['type'] != 'MX') {
+            return true;
+        }
+
+        // "Null MX" record indicates the domain accepts no mail (https://tools.ietf.org/html/rfc7505)
+        if (empty($dnsRecord['target']) || $dnsRecord['target'] == '.') {
+            $this->error = new DomainAcceptsNoMail();
+            return false;
+        }
+
+        $this->mxRecords[] = $dnsRecord;
 
         return true;
     }
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -98,7 +98,7 @@ class DNSCheckValidation implements EmailValidation
         $dnsRecords = dns_get_record($host, DNS_MX + DNS_A + DNS_AAAA);
 
         // No MX, A or AAAA records
-        if (empty($dnsRecords) || !is_array($dnsRecords)) {
+        if (empty($dnsRecords)) {
             $this->error = new NoDNSRecord();
             return false;
         }
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -4,6 +4,8 @@ namespace Egulias\EmailValidator\Validation;
 
 use Egulias\EmailValidator\EmailLexer;
 use Egulias\EmailValidator\Exception\InvalidEmail;
+use Egulias\EmailValidator\Exception\LocalOrReservedDomain;
+use Egulias\EmailValidator\Exception\DomainAcceptsNoMail;
 use Egulias\EmailValidator\Warning\NoDNSMXRecord;
 use Egulias\EmailValidator\Exception\NoDNSRecord;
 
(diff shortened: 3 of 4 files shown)


BenHarris:
Building now, but seems to be failing on PHP 5.5 with "Peer fingerprint did not match". Looks like a Travis CI issue?

src/Exception/LocalOrReservedDomain.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Egulias\EmailValidator\Exception;
+
+class LocalOrReservedDomain extends InvalidEmail
+{
+    const CODE = 153;
+    const REASON = 'Local, mDNS or reserved domain (RFC2606, RFC6762)';
```
egulias, line 8: 👍

src/Exception/DomainAcceptsNoMail.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Egulias\EmailValidator\Exception;
+
+class DomainAcceptsNoMail extends InvalidEmail
+{
+    const CODE = 154;
+    const REASON = 'Domain accepts no mail (Null MX, RFC7505)';
```
egulias, line 8: Is  `mail` correct or it should be `email` ?

src/Validation/DNSCheckValidation.php
```diff
@@ -36,6 +38,36 @@ public function isValid($email, EmailLexer $emailLexer)
...
+            'intranet',
+            'internal',
+            'private',
+            'corp',
+            'home',
+            'lan',
+        ];
+
+        // Exclude reserved top level DNS names
+        if (count($hostParts) <= 1 || in_array($hostParts[(count($hostParts) - 1)], $reservedTopLevelDnsNames)) {
```
egulias, line 66: If I understand well, 
`$hostParts[(count($hostParts) - 1)]` extracts the actual domain to check, right?
If so, would you mind extracting this to a local var for clarity?
After that you could go for `isset` instead of transversing the whole list if using the hosts as keys.

src/Validation/DNSCheckValidation.php
```diff
@@ -57,21 +89,46 @@ public function getWarnings()
...
+        if (empty($dnsRecords)) {
+            $this->error = new NoDNSRecord();
+            return false;
+        }
+
+        $aRecords  = [];
+        $mxRecords = [];
+
+        // Iterate over all returned DNS records
+        foreach ($dnsRecords as $dnsRecord) {
```
egulias, line 110: Please extract to a separate method.

tests/EmailValidator/Validation/DNSCheckValidationTest.php
```diff
@@ -14,50 +16,94 @@ public function validEmailsProvider()
     {
         return [
             // dot-atom
-            ['Abc@example.com'],
-            ['ABC@EXAMPLE.COM'],
-            ['Abc.123@example.com'],
-            ['user+mailbox/department=shipping@example.com'],
-            ['!#$%&\'*+-/=?^_`.{|}~@example.com'],
+            ['Abc@github.com'],
```
egulias, line 19: `example` is not a valid domain?

egulias:
Yes, it was Travis, working now.

src/Exception/DomainAcceptsNoMail.php
BenHarris: The title of RFC7505 is "A "Null MX" No Service Resource Record for Domains That Accept No Mail", so felt mail was more coherent.

src/Validation/DNSCheckValidation.php
BenHarris: It is two-fold

1. `count($hostParts) <= 1` to identify if it is a local domain (e.g. blah@localhost or blah@localdomain)
2. `in_array($hostParts[(count($hostParts) - 1)], $reservedTopLevelDnsNames)` is checking the if the tld is in the list of reserved top level dns names (e.g. .test, .example, .invalid etc.)

I think extracting it into a local domain is likely to make it more confusing. Could split it into two checks if you think it is really necessary.

tests/EmailValidator/Validation/DNSCheckValidationTest.php
BenHarris: It is a valid domain, but it also has a "Null MX" record, so now fails!

src/Validation/DNSCheckValidation.php
BenHarris: What is your reasoning for this? It's quite simple, so I feel that splitting it out will make it harder to read. If we weren't returning in it, I would agree, but returning there allows us to potentially save a few cpu cycles.

tests/EmailValidator/Validation/DNSCheckValidationTest.php
```diff
@@ -14,50 +16,94 @@ public function validEmailsProvider()
...
+    public function testDnsWarnings()
     {
         $validation = new DNSCheckValidation();
         $expectedWarnings = [NoDNSMXRecord::CODE => new NoDNSMXRecord()];
         $validation->isValid("example@invalid.example.com", new EmailLexer());
         $this->assertEquals($expectedWarnings, $validation->getWarnings());
     }
+    */
 
-    public function testNoDNSError()
```
BenHarris, line 56: Might be worth flowing this update through all of the tests (so they are all full camel case). It improves the display when you run the tests (e.g. you get `No dns error` instead of `No d n s error` when running PHPUnit)

src/Validation/DNSCheckValidation.php
```diff
@@ -57,21 +89,46 @@ public function getWarnings()
...
+                    $this->error = new DomainAcceptsNoMail();
+                    return false;
+                }
+
+                $mxRecords[] = $dnsRecord;
+                continue;
+            }
+
+            if (in_array($dnsRecord['type'], ['A', 'AAAA'])) {
+                $aRecords[] = $dnsRecord;
```
egulias, line 123: Whiy is ```$aRecords``` needed? I have failed to see usage of this var.

src/Validation/DNSCheckValidation.php
egulias: I see your point.
Main reasons are that there are 3 nesting levels, two main decisions branches and there is an important decision in the most inner one. So I'd say saving a couple of cycles is an acceptable price.
What I see is somehiting like
```php
//... L105

return $this->validateDNSRecords ($dnsRecords);
}

private function validateDNSRecords ($dnsRecords) {

    foreach ($dnsRecords as $dnsRecord) {
       if (!$this->validateMXRecord($dnsRecord)) { return false; }
       //add here other validations, i.e: AAA records?
    }
    if (empty($this->mxRecords)) {
            $this->warnings[NoDNSMXRecord::CODE] = new NoDNSMXRecord();
    }
    return true;
}
// returns boolean
private function validateMXRecord($dnsRecord) {
     if ($dnsRecord['type'] != 'MX') { return true; }
    // "Null MX" record indicates the domain accepts no mail (https://tools.ietf.org/html/rfc7505)
    if (empty($dnsRecord['target']) || $dnsRecord['target'] == '.') {
                    $this->error = new DomainAcceptsNoMail();
                    return false;
    }
    $this->mxRecords[] = $dnsRecord;
    return true;
}
```

tests/EmailValidator/Validation/DNSCheckValidationTest.php
egulias: However I'd like to find non-commercial examples. A quick search has failed me. I thought `mail.example.com` would work but it doesnt. Do you know any?

tests/EmailValidator/Validation/DNSCheckValidationTest.php
egulias: Give it a try, let see how it looks, in a separate PR please.

tests/EmailValidator/Validation/DNSCheckValidationTest.php
BenHarris: ietf.org, maybe?

tests/EmailValidator/Validation/DNSCheckValidationTest.php
BenHarris: Will do once we have finished this :-)

src/Validation/DNSCheckValidation.php
BenHarris: Not currently used. I was thinking about pinging or validating the target address(es). Technically, it could currently still be invalid/local. Probably getting towards overkill there though... I'll remove it.

src/Validation/DNSCheckValidation.php
BenHarris: I've split this out locally but I think it makes it significantly harder to understand what is going on at a glance. Ultimately your call though... Kinda feels like you only really want it for the single return in `checkDns()` haha

tests/EmailValidator/Validation/DNSCheckValidationTest.php
egulias: That would be perfect!

tests/EmailValidator/Validation/DNSCheckValidationTest.php
egulias: Supposedly is run by Scrutinizr, so run on every PR. Also Psalm. So the code should be widle analysed. Anyway, go for it. All improvements are welcome.

src/Validation/DNSCheckValidation.php
egulias: I think something got lost from the prvous comments, for this one.
My suggestion then would be something like this, to keep it simple yet more clear:
```php
$isLocalDomain = count($hostParts) <= 1;
$isReservedTopLevel = in_array($hostParts[(count($hostParts) - 1)], $reservedTopLevelDnsNames);
if ($isLocalDomain || isReservedTopLevel) {...}
```
What do you think?

src/Validation/DNSCheckValidation.php
egulias: I don't see any changes, have you pushed the commits?
Hahaha, if we could even remove that one return...  just kidding! What I'm looking for is to make it understandable by encapsulating, that ultimately did drive to a single return in the calling method, but as you can see on the private ones I have no issues with multiple returns.

Let see your changes and we can then solve this one!
Thanks!

BenHarris: 👍

src/Validation/DNSCheckValidation.php
BenHarris: Agreed

BenHarris:
Should be good to go now. Just a thought, but we should add psalms as a composer dev dependency.

src/Validation/DNSCheckValidation.php
```diff
@@ -36,7 +44,40 @@ public function isValid($email, EmailLexer $emailLexer)
...
+            'intranet',
+            'internal',
+            'private',
+            'corp',
+            'home',
+            'lan',
+        ];
+
+        $isLocalDomain = count($hostParts) <= 1;
+        $isReservedTopLevel = in_array($hostParts[(count($hostParts) - 1)], $reservedTopLevelDnsNames);
```
glensc, line 72: `
        $isReservedTopLevel = in_array($hostParts[(count($hostParts) - 1)], $reservedTopLevelDnsNames, true);
`


use strict mode by default

src/Validation/DNSCheckValidation.php
```diff
@@ -54,24 +95,74 @@ public function getWarnings()
...
+    /**
+     * Validate an MX record
+     *
+     * @param array $dnsRecord Given DNS record.
+     *
+     * @return bool True if valid.
+     */
+    private function validateMxRecord($dnsRecord)
+    {
+        if ($dnsRecord['type'] != 'MX') {
```
glensc, line 154: `        if ($dnsRecord['type'] !== 'MX') {
`

use strict mode

src/Validation/DNSCheckValidation.php
```diff
@@ -54,24 +95,74 @@ public function getWarnings()
...
+            return true;
+        }
+
+        // "Null MX" record indicates the domain accepts no mail (https://tools.ietf.org/html/rfc7505)
+        if (empty($dnsRecord['target']) || $dnsRecord['target'] == '.') {
```
glensc, line 159: `        if (empty($dnsRecord['target']) || $dnsRecord['target'] === '.') {
`

strict mode

egulias:
Approived. Thanks for your work @BenHarris !

PR #286, opened 2021-02-22
Use a safe way to execute dns_get_record (v3)

### LINKED ISSUE #256: dns_get_record(): DNS Query failed
```php
dns_get_record('test@t12est.co12m', DNS_MX + DNS_A + DNS_AAAA);
```

I got "Warning: dns_get_record(): DNS Query failed".

I suggest add @ to dns_get_record() in https://github.com/egulias/EmailValidator/blob/840d5603eb84cc81a6a0382adac3293e57c1c64c/src/Validation/DNSCheckValidation.php#L121

My environment:
- Windows 10
- PHP 7.3.12 (cli) (built: Nov 19 2019 13:58:02) ( ZTS MSVC15 (Visual C++ 2017) x64 )

alies-dev:
Closes #256 (v3 branch)

The idea is to create a custom error handler for `dns_get_record()`, call `dns_get_record` and then restore the original error handler.

This way we can catch PHP warnings (e.g. `Warning: dns_get_record(): DNS Query failed`) and errors: `dns_get_record(): A temporary server error occurred.` [ErrorException].

a new `UnableToGetDNSRecord` extends `NoDNSRecord` for backward compatibility reasons. Note sure about `CODE` const at the exception, I haven't found any rules for it.

## Why @ should be avoided

See https://derickrethans.nl/five-reasons-why-the-shutop-operator-should-be-avoided.html

In short:
 - can produce annoying side effects
 - it's slow

## Why a new test disabled by default
`dns_get_record` returns false, generates E_WARNING in edge cases, that we can't control or easily mock from PHP code (e.g. timeouts). when use `dns_get_record($host, \DNS_ALL)`, it takes 60secs on my OS (macOS), 30secs on Docker container (ubuntu 18.04-based). For this reason I've decided to exclude this test by default on CI.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -7,7 +7,7 @@ build:
     tests:
         override:
             -
-                command: 'vendor/bin/phpunit --coverage-clover=clover.xml'
+                command: 'vendor/bin/phpunit --coverage-clover=clover.xml --exclude-group slow'
                 coverage:
                     file: 'clover.xml'
                     format: 'clover'
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_script:
   - mkdir -p build/logs
 
 script:
-  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml
+  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml --exclude-group slow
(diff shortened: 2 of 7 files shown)


egulias:
Thanks @lptn !

PR #287, opened 2021-02-22
Use a safe way to execute dns_get_record (v2) 

alies-dev:
Backport of the `Use a safe way to execute dns_get_record (v3)` #286
___________

The idea is to create a custom error handler for dns_get_record(), call dns_get_record and then restore the original error handler.
This way we can catch PHP warnings (e.g. "Warning: dns_get_record(): DNS Query failed" or "dns_get_record(): A temporary server error occurred." [ErrorException])

## Tests

About failed tests: I think it’s not easily possible to test this case. Probably we can even remove a test I've added.

Also, I can't change Scrutinizer, it still tries to execute all tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ before_script:
   - mkdir -p build/logs
 
 script:
-  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml
+  - composer test:ci -- --coverage-clover build/logs/clover.xml
   - if [ "$psalm" = "yes" ]; then vendor/bin/psalm; fi
 
 after_script:
--- a/composer.json
+++ b/composer.json
@@ -34,5 +34,9 @@
     "psr-4": {
       "Egulias\\EmailValidator\\Tests\\": "tests"
     }
+  },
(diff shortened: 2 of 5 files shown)


egulias:
Looks like everything above PHP7 is failing due to psalm errors.

alies-dev:
@egulias 

>Looks like everything above PHP7 is failing due to psalm errors.

exactly, but we can't use parameter types bc on v2 we support PHP >=5.5

e.g.:
MissingClosureParamType - src/Validation/DNSCheckValidation.php:128:30 - Parameter $errorLevel has no provided type (see https://psalm.dev/153)

I would say it's psalm configuration issue, but I've just added suppression statement to "resolve" it: 06f4e8f

alies-dev:
Hey @egulias

what is the state of this PR review? Do you want something extra from me as from PR author?

egulias:
Hi @lptn , thanks for the heads up. I have had no time to check this one. I will review it shortly. 
I'll ping you if anything else is needed.

egulias:
Sorry, won't merge.

PR #297, opened 2021-03-24
Honor RFC 5231 Domain Length

johncongdon:
According to RFC 5321 The Domain Part can be 255 octets.

Since this is testing just the domain part, we shouldn't be checking on the local label length of 64.

4.5.3.1.1.  Local-part
   The maximum total length of a user name or other local-part is 64 octets.

4.5.3.1.2.  Domain
   The maximum total length of a domain name or number is 255 octets.

https://tools.ietf.org/html/rfc5321

I moved some of the test data around to make it match correctly based on these rules.

Twitter: @johncongdon

### DIFF
--- a/src/Parser/DomainPart.php
+++ b/src/Parser/DomainPart.php
@@ -280,9 +280,9 @@ class DomainPart extends PartParser
     {
         if ($this->lexer->token['type'] === EmailLexer::S_DOT || $isEndOfDomain) {
             if ($this->isLabelTooLong($this->label)) {
-                $this->label = '';
                 return new InvalidEmail(new LabelTooLong(), $this->lexer->token['value']);
             }
+            $this->label = '';
         }
         $this->label .= $this->lexer->token['value'];
         return new ValidEmail();
--- a/tests/EmailValidator/Validation/RFCValidationDomainPartTest.php
+++ b/tests/EmailValidator/Validation/RFCValidationDomainPartTest.php
@@ -76,6 +76,7 @@ class RFCValidationDomainPartTest extends TestCase
             ['validipv4@[127.0.0.0]'],
             ['validipv4@127.0.0.0'],
             ['withhyphen@domain-exam.com'],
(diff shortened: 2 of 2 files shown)


egulias:
Hi @johncongdon , thanks for your PR.
Domain part intends to validate against RFC1035 as RFC5321 was too open and not implemented in reality for domains. In RFC1035 (as far as I have understood) a `label` is the string between dots andthey can have a maximum size of 63 each (https://tools.ietf.org/html/rfc1035#section-2.3.4), while the whole domain name a max of 255.
So, these
```php
['example@toolonglocalparttoolonglocalparttoolonglocalparttoolonglocalpart.co.uk'],
            ['example@toolonglocalparttoolonglocalparttoolonglocalparttoolonglocalpart.test.co.uk'],
            ['example@test.toolonglocalparttoolonglocalparttoolonglocalparttoolonglocalpart.co.uk'],
```

should continue to be invalid because `toolonglocalparttoolonglocalparttoolonglocalparttoolonglocalpart` is 64 chars. Same for 
`[new InvalidEmail(new DomainTooLong(), ""), sprintf('example@%s.com', str_repeat('ъ', 255))],`

Now, it is true that to really validate that  it works properly a new test is needed, something like 
` [new InvalidEmail(new DomainTooLong(), ""), sprintf('example@%s.%s.%s.%s.com', str_repeat('a', 63))],`
to have valid labels and an invalid domain.

johncongdon:
I'll have to read through, but either way, I have an email address from salesforce that violated that which lead me to creating this pr.

With this I am able to email my salesforce email address.

johncongdon:
@egulias thank you for the lesson.  I think I did track down the actual bug now.

When looping through the parts, we are not resetting the label to empty when encountering a S_DOT.  So the label keeps growing and ends up counting all of the characters in the domain.

I've included a DOMAIN in the test that is VERY similar to the sales force one I mentioned earlier.

johncongdon:
BTW: Did that as a TDD exercise.  Added the new test email address in the tests and watched it fail.  Spent a ton of time tracing the code via xdebug until I understood what the code was doing and how to fix it.  Test passed.  :tada:

egulias:
Well done @johncongdon , all green! Merging.
Thanks for your contribution.

PR #304, opened 2021-09-24
MultipleErrors fails when retrieving the first reason, when this has a reason code <> 0

samuelvi:
Having this adder method:

    public function addReason(Reason $reason) : void
    {
        $this->reasons[$reason->code()] = $reason;
    }

Will fail when

    public function reason() : Reason
    {
        return $this->reasons[0];
    }

If the $reason->code is different to 0.

That's why I propose in addReason having this line instead: $this->reasons[] = $reason;

### DIFF
--- a/src/Result/MultipleErrors.php
+++ b/src/Result/MultipleErrors.php
@@ -21,7 +21,7 @@ class MultipleErrors extends InvalidEmail
 
     public function addReason(Reason $reason) : void
     {
-        $this->reasons[$reason->code()] = $reason;
+        $this->reasons[] = $reason;
     }
 
     /**


egulias:
Hello @samuelvi thanks for your PR, well spotted.
Can you please add a test for the case?
The intention of the code is to store reasons by their code or ease of access. Your change fixes the bug and removes that.
Maybe you could find a way for the accessor method `reason()` to return the first element of the array no matter the index.

samuelvi:
Hi,
I'll try to test it asap, quite load of work just now, but I'll try for sure.
We found a workaround for getting the first index, but as fas as I understand is a temp fix.
Thanks a lot!

samuelvi:
Hi,
I created a new pull request: https://github.com/egulias/EmailValidator/pull/305
Sorry if that confuses you.
Thanks.

PR #305, opened 2021-10-01
MultipleErrors->reason always returns [0] and should return first item from associative array

samuelvi:
Hi,
Adapted the MultipleErrors code for getting the first reason from array, instead of the one from [0].
As $this->reasons is an associative array, first item not always equals 0, it depends on the $reason->code() passed to the 
 addReason method.
Created a new custom Exception when $this->reasons is empty.
Created 3 more test methods into the file MultipleErrorsTest.


egulias:
Hi @samuelvi , nice PR, thank you.
Throwing and exception on an accesor method is not advisable as you are upseting the "consumer" flow, now one expects it. Moreover when the user has no control on why is that happening (that is, user is not setting/unsetting reasons).
Please change the use of an exception to an `EmptyReason`. This way you can return object, which will also typecast to `Reason`, and won't disturb userland flow.

Thanks!

samuelvi:
Hi,
I adapted the code as you mentioned me, please if there is something I did not understand correctly don't hesitate to let me know.
Regarding tests, when I run locally always get an error on DNSCheckValidationTest::testDomainAcceptsNoMailError, but if I run only the MultipleErrorsTest.php (./vendor/phpunit/phpunit/phpunit tests/EmailValidator/Reason/MultipleErrorsTest.php) passes correctly.
I hope it helps you somehow.
Thanks a lot!

egulias:
Looks great @samuelvi , well done. Yes, is a flaky test, I have to check what's going on there.
Thanks for your PR, merging!

PR #306, opened 2021-10-17
Small contribution - code readability improvements

JuGid:
Hi,

Thanks for your code !

As a very small contribution to this library, I would like to propose some code readability/maintainability improvements. This will not revolutionize anything, but hope this is good enough.

I tried to run tests but there was some errors on DSNCheckValidationTest. This error was before and still returned by PHPUnit. May be linked to #287 ?

```
1) Egulias\EmailValidator\Tests\EmailValidator\Validation\DNSCheckValidationTest::testNoDNSError
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 Egulias\EmailValidator\Result\InvalidEmail Object (
     'token' => ''
-    'reason' => Egulias\EmailValidator\Result\Reason\NoDNSRecord Object &0000000000d01c93000000002db9369d ()
+    'reason' => Egulias\EmailValidator\Result\Reason\UnableToGetDNSRecord Object &0000000000d01c94000000002db9369d ()
 )
```

Hope this will not block this PR. This PR doesn't change EmailValidator functionally.

Regards,
Julien

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -7,7 +7,7 @@ build:
     tests:
         override:
             -
-                command: 'vendor/bin/phpunit --coverage-clover=clover.xml --exclude-group slow'
+                command: 'vendor/bin/phpunit --coverage-clover=clover.xml --exclude-group flaky'
                 coverage:
                     file: 'clover.xml'
                     format: 'clover'
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
   - mkdir -p build/logs
 
 script:
-  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml --exclude-group slow
+  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml --exclude-group flaky
   - if [ "$psalm" = "yes" ]; then vendor/bin/psalm; fi
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
   - mkdir -p build/logs
 
 script:
-  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml --exclude-group flaky
+  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml --exclude-group slow
   - if [ "$psalm" = "yes" ]; then vendor/bin/psalm; fi
 
 after_script:
--- a/src/EmailLexer.php
+++ b/src/EmailLexer.php
@@ -211,8 +211,12 @@ class EmailLexer extends AbstractLexer
         }
 
         $this->previous = $this->token;
+        
+        if($this->lookahead === null) {
+            $this->lookahead = self::$nullToken;
+        }
+
         $hasNext = parent::moveNext();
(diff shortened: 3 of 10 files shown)


egulias:
Hi @JuGid ,
Thanks for your work! Looks good to me, well done.
The test is failing because there's a yaml line to update in https://github.com/egulias/EmailValidator/blob/3.x/.scrutinizer.yml#L10 for a test group name change.
Do you mind fixing that in this PR?
Thanks!

JuGid:
Yes, sure ! On duty !

EDIT : To be honest, I don't really know what to change now compared to #287 and Travis fails now

egulias:
Hi @JuGid ,
I request you to change the file .scrutinizer.yml in line 10 to update to `flaky`. You have changed .travis.yml to `slow`.
That's why Travis is failing. 
So, please

- you have to revert your change in travis.yml
- change https://github.com/egulias/EmailValidator/blob/3.x/.scrutinizer.yml#L10 to `flaky`

Travis is also failing for psalm errors, aparently new. Sorry for that. You can run psalm in local to see what is complainng about.

JuGid:
I don't know why psalm fail since $token has types array|nul and is already set in constructor but moveNext() can set $token to null from lookahead variable (AbstractLexer)l. I'm running it with php 8.0.

```
ERROR: RedundantPropertyInitializationCheck - src/EmailLexer.php:215:24 - Property $this->token with type array{position: int, type: int|null, value: string} should already be set in the constructor (see https://psalm.dev/261)
        $this->token = $this->token ?? self::$nullToken;

ERROR: RedundantPropertyInitializationCheck - src/EmailLexer.php:215:40 - Property $this->token with type array{position: int, type: int|null, value: string} should already be set in the constructor (see https://psalm.dev/261)
        $this->token = $this->token ?? self::$nullToken;
```
May be this is caused by this ratio :` Psalm was able to infer types for 99.5614% of the codebase`, don't know...

**Work in progress :)**

egulias:
Thanks for the effort @JuGid. 
Reading the error, my thoughts are:
The first error you got is that `?:` was always to be true. This might be because `$this->token` is an array and upon evaluation it will never evaluate to false.
The code is validating if there is a token set or else assign the null token. To fix the issue, keep `?:` operator and improve the `$this->token ?: self::$nullToken;` evaluation. For instance, `empty ($this->token)` might do the trick.
Other option is to create a test prooving that that line fails for an empty token.

JuGid:
I found out that there is a problem since $token is defined in AbstractLexer (array|null) and in EmailLexer(array). In the script, parent::moveNext is used causing a big misunderstanding (for me). When you use parent::moveNext(), $this->lookahead can be null and so, the token. So I just verified if lookahead is null, then set it to nullToken and moveNext to have a token never null. Now psalm and php-unit are good with it. 

Thank you for the small tips you gave me and allowed me to go forward :) 

EDIT : btw I learnt a lot about PSALM ! That was insanely cool !

egulias:
Hi @JuGid, well done! Thanks for your time and effort. And sorry for my delay. Merging!

PR #325, opened 2022-06-16
Don't use utf8_en/decode() functions, they're deprecated on PHP 8.2

nicolas-grekas:
For the call to idn_to_ascii(), [the doc](https://php.net/idn_to_ascii) tells that an UTF-8 string must be passed, so that calling utf8_decode() before doesn't make sense to me. I just removed the call.


egulias:
Thanks!

PR #327, opened 2022-07-04
Final public method for abstract class

MathiasReker:
All public methods of abstract classes should be final. Enforce API encapsulation in an inheritance architecture. If you want to override a method, use the Template method pattern.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -32,7 +32,7 @@ abstract class Parser
         $this->lexer = $lexer;   
     }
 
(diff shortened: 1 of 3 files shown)


egulias:
Hi @MathiasReker. Thanks for your PR.
You PR is not passing the tests because it requires a rewrite of several classes inheriting from the ones you are modifying.
These methods are accessor methods to information (e.g get the warnings), rather than a behaviour itself. This pattern is used in other places were appropriate.

Won't merge this one.

PR #335, opened 2022-09-09
Auto multi validation

mariuszmalek:
I created this branch to have every validation in one place by class name.

Example use:

```
$array = EmailValidatorFactory::create("my@email.com");
```

### DIFF
--- a/src/EmailValidatorFactory.php
+++ b/src/EmailValidatorFactory.php
@@ -4,9 +4,8 @@ declare(strict_types=1);
 
 namespace Egulias\EmailValidator;
 
+use Egulias\EmailValidator\EmailValidator;
 use Egulias\EmailValidator\Validation\DNSCheckValidation;
-use Egulias\EmailValidator\Validation\DNSGetRecordWrapper;
-use Egulias\EmailValidator\Validation\MessageIDValidation;
 use Egulias\EmailValidator\Validation\NoRFCWarningsValidation;
 use Egulias\EmailValidator\Validation\RFCValidation;
 
@@ -16,8 +15,6 @@ class EmailValidatorFactory
     protected static array $defaultValidators = [
         RFCValidation::class,
         NoRFCWarningsValidation::class,
-        MessageIDValidation::class,
-        DNSGetRecordWrapper::class,
         DNSCheckValidation::class
     ];
     
--- a/src/EmailValidatorFactory.php
+++ b/src/EmailValidatorFactory.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Egulias\EmailValidator;
+
+use Egulias\EmailValidator\Validation\DNSCheckValidation;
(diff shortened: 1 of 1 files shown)


mariuszmalek:
Thanks to this, I can now easily handle all the information and save it in the database.

src/EmailValidatorFactory.php
```diff
@@ -0,0 +1,36 @@
...
+        RFCValidation::class,
+        NoRFCWarningsValidation::class,
+        DNSCheckValidation::class
+    ];
+    
+    /**
+     * @param string $emailAddress
+     * @return array
+     */
+    public static function create(string $emailAddress): array
```
egulias, line 25: Which are the reasons behing making it static? It prevents being injected as dependency.

src/EmailValidatorFactory.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $validator = new EmailValidator();
+        $result = [];
+
+        foreach (self::$defaultValidators as $key => $val) {
```
egulias, line 30: You could just use ´MultipleValidationWithAnd´ at construction time. And remove line 15.

src/EmailValidatorFactory.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Egulias\EmailValidator\EmailValidator;
+use Egulias\EmailValidator\Validation\DNSCheckValidation;
+use Egulias\EmailValidator\Validation\NoRFCWarningsValidation;
+use Egulias\EmailValidator\Validation\RFCValidation;
+
+class EmailValidatorFactory
+{
+    /** @var [] */
+    protected static array $defaultValidators = [
```
egulias, line 15: Are you expecting or encouraging extension? This prevents it. There is also no entry point to add additional/other validators than this.

You can do it via constructor arguments with sensible defaults (is like your idea here, but with simple extension point)

src/EmailValidatorFactory.php
egulias: Also, you are returning the lists of results, not a validator ready for use. I'd suggest to change the name to something like `validationsResults` or similar.

src/EmailValidatorFactory.php
```diff
@@ -0,0 +1,36 @@
...
+    public static function create(string $emailAddress): array
+    {
+        $validator = new EmailValidator();
+        $result = [];
+
+        foreach (self::$defaultValidators as $key => $val) {
+            $result[get_class(new $val)] = $validator->isValid($emailAddress, new $val);
+        }
+
+        return $result;
```
egulias, line 34: Is the validator's `Result` type not useful? Maybe adding which validation is giving which result would be useful?

egulias (changes requested):
Hey. Thanks for your patience. Leaving you here some comments.

egulias:
Closing. Branch 3.x is no longer maintained.

PR #339, opened 2022-12-14
Migrate to doctrine/lexer v3.0

Ph0tonic:
[Doctrine/L](https://github.com/doctrine/lexer) just released a brand new version 3.0 and here is a PR to use this newer version.

Let me know @egulias if I need to change anything, I am not familiar with psalm annotations.

PR #340, opened 2022-12-20
Allow doctrine/lexer 2

derrabus:
This is a lightweight version of #339 that allows `doctrine/lexer` 2 to be used while maintaining backwards compatibility.

For Symfony's CI, we need this change because EmailValidator currently blocks the installation of `doctrine/annotations` 2 (which itself depends on Lexer 2).

This PR by no means intends to replace #339: #339 is still useful because it fully migrates this project to Lexer's new API. But since it involves BC breaks, I suggest to merge this PR into 3.x and release #339 as v4.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,13 +14,13 @@
   },
   "require": {
     "php": ">=7.2",
-    "doctrine/lexer": "^1.2",
+    "doctrine/lexer": "^1.2|^2",
     "symfony/polyfill-intl-idn": "^1.15"
   },
   "require-dev": {
     "php-coveralls/php-coveralls": "^2.2",
     "phpunit/phpunit": "^8.5.8|^9.3.3",
-    "vimeo/psalm": "^4"
+    "vimeo/psalm": "^4.30|^5.4"
   },
   "suggest": {
     "ext-intl": "PHP Internationalization Libraries are required to use the SpoofChecking validation"
--- a/psalm.xml
+++ b/psalm.xml
@@ -3,7 +3,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config ./vendor/vimeo/psalm/config.xsd"
-    errorBaseline="./psalm.baseline.xml"
+    errorBaseline="psalm.baseline.xml"
 >
     <projectFiles>
         <directory name="src" />
@@ -13,5 +13,11 @@
     </projectFiles>
 
     <issueHandlers>
+        <DeprecatedMethod>
+            <errorLevel type="suppress">
+                <!-- This issue needs to be resolved before upgrading to Lexer 3 -->
+                <referencedMethod name="Doctrine\Common\Lexer\Token::offsetGet"/>
+            </errorLevel>
+        </DeprecatedMethod>
     </issueHandlers>
 </psalm>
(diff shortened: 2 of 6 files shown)


src/EmailLexer.php
```diff
@@ -140,18 +144,20 @@ class EmailLexer extends AbstractLexer
...
-     * @psalm-suppress NonInvariantDocblockPropertyType
+     * @psalm-var array{value:string, type:null|int, position:int}|Token<int, string>
      */
     public $token;
 
     /**
      * The next token in the input.
      *
-     * @var array{position: int, type: int|null|string, value: int|string}|null
+     * @var array|Token
```
stof: changing the type of a public property in a non-internal class is still a BC break.

derrabus: It's an inherited BC break. This public property is declared in the Lexer library and apparently just overridden here to make the static analysis happy. As long as you pin Lexer to v1 in your project, the type won't change for you.

derrabus:
The Travis failure is unrelated and seems to be a hiccup on the build runner.

psalm.xml
```diff
@@ -13,5 +13,11 @@
     </projectFiles>
 
     <issueHandlers>
+        <DeprecatedMethod>
+            <errorLevel type="suppress">
+                <!-- This issue needs to be resolved before upgrading to Lexer 3 -->
+                <referencedMethod name="Doctrine\Common\Lexer\Token::offsetGet"/>
```
egulias, line 19: Which is the issue here? Can you reference to where in the code?

composer.json
```diff
@@ -14,13 +14,13 @@
   },
   "require": {
     "php": ">=7.2",
-    "doctrine/lexer": "^1.2",
+    "doctrine/lexer": "^1.2|^2",
```
egulias, line 17: Is there a reason why not to require v2 only?

egulias (changes requested):
Hey, apologies for the delay.
Can you please review all the Scrutinizr commnets? Seems there is a type miss-match.
And make v2 the only accepcted version.
Thus, you will need to change all the types to `Token` I believe.

psalm.xml
derrabus: In Lexer 1, tokens were associative arrays. Lexer 2 turns them into objects by introducing a new class `Token`. In order to make the migration easier, the new Token class implements `ArrayAccess` so that calls like this one are possible in both versions:

https://github.com/egulias/EmailValidator/blob/0170967656f55a87054fde439ef35a22df91468b/src/EmailLexer.php#L210

Avoiding the deprecation would mean, going from `$this->token['value']` to `$this->token->value` which is what #339 does already.

composer.json
derrabus: Lexer is a transitive dependency in many projects, shared by more than one library. Dropping v1 right away might cause conflicts downstream at the moment.

derrabus:
> Can you please review all the Scrutinizr commnets? Seems there is a type miss-match.

Scrutinizer apparently does not understand generics at all. Not sure if I can do anything about it. This part should be well covered by Psalm anyway.

egulias:
Checking Doctrine Lexer's code, it should work so we can ignore Scrutinizr, which in any case are informational anyway.

derrabus:
@egulias The PR was merged, but somehow one of your later merges seems to have reverted my changes. Was that on purpose?

egulias:
I merged this one and #343 , I might have resolved incorrectly the conflicts.

derrabus:
I've re-submitted this PR as #345.

PR #345, opened 2022-12-29
Allow doctrine/lexer 2

derrabus:
This is #340 cherry-picked on 3.x. The changes of #340 apparently got lost in a failed merge. Let's try this again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   },
   "require": {
     "php": ">=7.2",
-    "doctrine/lexer": "^1.2",
+    "doctrine/lexer": "^1.2|^2",
     "symfony/polyfill-intl-idn": "^1.15"
   },
   "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,35 +4,80 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a77d36b64bc1213fecf4d4f92d759c3b",
+    "content-hash": "9b420b2279da45fe448cd428cf568d87",
     "packages": [
         {
-            "name": "doctrine/lexer",
-            "version": "1.2.3",
+            "name": "doctrine/deprecations",
+            "version": "v1.0.0",
             "source": {
                 "type": "git",
-                "url": "https://github.com/doctrine/lexer.git",
-                "reference": "c268e882d4dbdd85e36e4ad69e02dc284f89d229"
+                "url": "https://github.com/doctrine/deprecations.git",
+                "reference": "0e2a4f1f8cdfc7a92ec3b01c9334898c806b30de"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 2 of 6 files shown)


egulias:
The problem is PHP 7.3, there are dependencies requiring PHP 7.4 and above. 
From Travis (https://app.travis-ci.com/github/egulias/EmailValidator/jobs/592229284 )
Problem 1

    - composer/pcre is locked to version 3.1.0 and an update of this package was not requested.

    - composer/pcre 3.1.0 requires php ^7.4 || ^8.0 -> your php version (7.3.14) does not satisfy that requirement.

  Problem 2

    - phpdocumentor/type-resolver is locked to version 1.6.2 and an update of this package was not requested.

    - phpdocumentor/type-resolver 1.6.2 requires php ^7.4 || ^8.0 -> your php version (7.3.14) does not satisfy that requirement.

  Problem 3

    - psr/container is locked to version 1.1.2 and an update of this package was not requested.

    - psr/container 1.1.2 requires php >=7.4.0 -> your php version (7.3.14) does not satisfy that requirement.

  Problem 4

    - composer/pcre 3.1.0 requires php ^7.4 || ^8.0 -> your php version (7.3.14) does not satisfy that requirement.

    - composer/xdebug-handler 3.0.3 requires composer/pcre ^1 || ^2 || ^3 -> satisfiable by composer/pcre[3.1.0].

    - composer/xdebug-handler is locked to version 3.0.3 and an update of this package was not requested.


Checking Symfony as a guideline, PHP 7.2 was there up until v5.4.
Maybe is Lexer v2 bringing some other requirements?

derrabus:
`composer/pcre` is already locked at that version on the 3.x branch. This PR does not change that.

egulias:
Last build before #340 was #333 and it was all good. So something has changed in the latest lock.

egulias:
It is not only `composer/pcre` and they were all locked.

egulias:
Ja!
PHP 7.3 was not in Travis matrix before 🤔

derrabus:
I mean, for a library with such a broad PHP version range, committing a lockfile doesn't really make sense. You almost certainly need different sets of dependencies for PHP 7.3 and 8.2.

I'll try to generate a compatible lockfile for PHP 7.3 in a separate PR.

egulias:
@derrabus I don't think is needed to support PHP 7.3, this seems to be an error from the merge. Just remove PHP 7.3 from `.travis.yml`

However your comment makes sense. Would you remove the `.lock` file in v4, for instance?

derrabus:
> Just remove PHP 7.3 from `.travis.yml`

That won't be enough: Scrutinizer fails because of the same issue. See #346 for a (hopefully) compatible lockfile.

> However your comment makes sense. Would you remove the `.lock` file in v4, for instance?

You can remove the lockfile anytime you want. It has no consequences for consumers of your library. It only affects which dependencies are installed in your CI and for local development when working on this repository.

egulias:
I need a fresh mind, will come back tomorrow (earlier, hopefully).

derrabus:
Sure thing. I should get some sleep as well. 😅

PR #346, opened 2022-12-29
Regenerate lockfile for PHP 7.3

derrabus:
🤞🏻


egulias:
Man, you are fast!

derrabus:
I've also regenerated the Psalm baseline because it didn't match the locked Psalm version anymore. That should (hopefully) get us a green CI.

derrabus:
All right. Apart from a hanging Travis runner that I cannot really fix, the CI is back to green now.

derrabus:
After 3.x has been reset, I don't think we still need this PR. Closing.

Issue #347, opened 2022-12-30
Several changes did not make it in 3.2.2 (Bad merge?)


DavidPrevot:
Some changes were reverted before the last release, and are not in the 3.x branch anymore. Simple examples: 
4c72190f321eaf59f690a5b848707044a73da858 (Update LICENSE) and “Allow doctrine/lexer 2 by @derrabus in https://github.com/egulias/EmailValidator/pull/340”.

Regards

derrabus:
Thanks. @egulias and I are already working on restoring those changes.

sbuerk:
Also the utf8 en/decode removal (PHP8.2 deprecation) are missing in 3.2.2 -> https://github.com/egulias/EmailValidator/commit/f88dcf4b14af14a98ad96b14b2b317969eab6715

Just FYI. Thanks for taking care @derrabus 

-> https://git.typo3.org/typo3/CI/cms/-/jobs/1847583/raw

driesvints:
It seems after the 3.2.2 release quite a few of our package's test suites started to break with the "hello@example.com" email address: 

https://github.com/laravel/telescope/actions/runs/3804233902/jobs/6471295811
https://github.com/laravel/framework/actions/runs/3804298612/jobs/6473840109

Any idea what change caused this?

manuelkiessling:
@derrabus Please let us know if you need help from the community to collect all commits that went missing in action.

driesvints:
The diff from 3.2.1 and 3.2.2 shows the changes which are the culprit for the issues here but I cannot seem to pinpoint the commits: https://github.com/egulias/EmailValidator/compare/3.2.1...3.2.2

manuelkiessling:
> The diff from 3.2.1 and 3.2.2 shows the changes which are the culprit for the issues here but I cannot seem to pinpoint the commits: [3.2.1...3.2.2](https://github.com/egulias/EmailValidator/compare/3.2.1...3.2.2)

Yeah it's super-weird and I start to actually question my understanding of how Git works - but there simply are no commits that e.g. revert the `utf8_encode` line, and yet it's gone. Also, CLI `git log src/EmailLexer.php` on current 3.x branch only shows commits up to 007bfd4 inclusive, omitting the 4 commits a5ed8d58, f88dcf4b, 18b9e5b5, and b28de872 that followed it. However, IntelliJ Git Log UI does show them (but their changes are nevertheless not in the file).

manuelkiessling:
[...]

PR #348, opened 2022-12-30
GitHub Actions

driesvints:
This PR adds new GitHub Actions build that will replace the old Travis one. I left comments on the most significant parts I changed below.


.coveralls.yml
```diff
@@ -1,3 +1,3 @@
-service_name: travis-ci
+service_name: github-actions
```
driesvints: I'm not 100% about this change. Obviously we won't be using Travis anymore. Coveralls mentions this can be "any string" but only documents travis-ci or travis-pro to be proper values. So unsure what to do here.

https://github.com/php-coveralls/php-coveralls#coverallsyml

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
+name: tests
+
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
```
driesvints, line 10: This will run your build one time during the day to spot regressions in third party libraries.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  tests:
+    runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: true
+      matrix:
+        php: [7.4, '8.0', 8.1, 8.2]
```
driesvints: I only tested from 7.4 and up since that was the lowest version in Travis.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+          extensions: dom, curl, libxml, mbstring, redis, zip
+          ini-values: error_reporting=E_ALL
+          tools: composer:v2
+          coverage: none
+
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-interaction --no-progress
+
+      - name: Execute Psalm
+        run: vendor/bin/psalm
```
driesvints: Running Psalm first because it has a lower lifecycle to run its checks than PHPUnit.

.gitignore
```diff
@@ -1,6 +1,8 @@
-phpunit.result.cache
-.idea
+build/
 report/
 vendor/
+composer.lock
```
driesvints, line 4: I removed the composer.lock file from this repo. I'm not sure why it was committed? A composer.lock file is generated with environment info at the time dependencies were installed (like the PHP version). It's unsuited to be committed to be run as part of a CI build. I don't think it should stay hardcoded in a package repo.

phpunit.xml.dist
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="false" backupStaticAttributes="false" bootstrap="vendor/autoload.php" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" executionOrder="defects" processIsolation="false" stopOnFailure="false">
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="false" backupStaticAttributes="false" bootstrap="vendor/autoload.php" colors="true" convertDeprecationsToExceptions="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" executionOrder="defects" processIsolation="false" stopOnFailure="false">
```
driesvints, line 2: Converting deprecations to exceptions will help uncover issues like the `utf8_encode` one in https://github.com/egulias/EmailValidator/issues/347

driesvints:
@derrabus @egulias this one is ready now. Feel free to ask questions if you have some.

.github/workflows/tests.yml
derrabus: Wouldn't it make sense to move Psalm into a separate workflow? I would want to run PHPUnit even if Psalm fails. Additionally, both tools can perfectly run in parallel.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: dom, curl, libxml, mbstring, redis, zip
+          ini-values: error_reporting=E_ALL
+          tools: composer:v2
+          coverage: none
+
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-interaction --no-progress
```
derrabus: Let's leverage https://github.com/ramsey/composer-install ?

.github/workflows/tests.yml
driesvints: We can definitely do that.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: dom, curl, libxml, mbstring, redis, zip
```
derrabus: What do we need all those extensions for?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+          ini-values: error_reporting=E_ALL
+          tools: composer:v2
+          coverage: none
```
derrabus: PHPUnit is called with options that generate a coverage report. Let's install Xdebug or pcov?

.github/workflows/tests.yml
driesvints: Done.

.github/workflows/tests.yml
derrabus: We could try to add 7.3 and 7.2 to the matrix though and keep them if they run green.

.github/workflows/tests.yml
driesvints: I copy/pasted a different workflow but you're right, probably not needed.

.github/workflows/tests.yml
driesvints: Good catch!

.github/workflows/tests.yml
driesvints: Ah, never heard of it. Thanks for pointing it out.

.github/workflows/tests.yml
driesvints: Seems to work!

derrabus:
Thanks, I like it! I don't know how to connect Coveralls either, though.

egulias:
Coveralls is/was used for code coverage, but that is also embedded into Scrutinizer which can in turn also be configured to add failure condition on coverage reduction.
We can remove coveralls al together, it is not providing value right now.

.coveralls.yml
egulias: Can you please remove coveralls (dependency and config file)? Taking the opportunity for some clean-up :)

driesvints:
@egulias I removed it 👍

egulias:
Thanks so much @driesvints !

driesvints:
No worries. We've been using `egulias/EmailValidator` for a while now so happy to give a bit back 🙂

PR #349, opened 2023-01-02
Fix Lexer 2 compatibility

derrabus:
This PR fixes an incompatibility with Lexer 2 due to an unnecessarily overridden property.

More details in the comments below.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,9 +16,14 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4, '8.0', 8.1, 8.2]
+        php: ['8.1', '8.2']
+        deps: [highest]
 
-    name: PHP ${{ matrix.php }}
+        include:
+          - php: '8.1'
+            deps: lowest
+
+    name: PHP ${{ matrix.php }} / ${{ matrix.deps }}
 
     steps:
       - name: Checkout code
@@ -34,6 +39,8 @@ jobs:
 
       - name: Install dependencies
         uses: ramsey/composer-install@v2
+        with:
+          dependency-versions: ${{ matrix.deps }}
 
       - name: Setup logs directory
         run: mkdir -p build/logs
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -16,9 +16,14 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4, '8.0', 8.1, 8.2]
```
derrabus, line 19: PHP 7.2-8.0 are not supported on this branch, so let's not test against them.

.github/workflows/tests.yml
```diff
@@ -16,9 +16,14 @@ jobs:
...
+        php: ['8.1', '8.2']
+        deps: [highest]
 
-    name: PHP ${{ matrix.php }}
+        include:
+          - php: '8.1'
+            deps: lowest
```
derrabus, line 24: This new Job tests with low dependencies. This way, issues like the one I've just fixed can be spotted more easily.

src/EmailLexer.php
```diff
@@ -144,13 +144,6 @@ class EmailLexer extends AbstractLexer
      */
     public Token $current;
 
-    /**
-     * The next token in the input.
-     *
-     * @var Token<int, string>|null
-     */
-    public Token|null $lookahead;
```
derrabus, line 152: This property exists on the parent class and we don't need to override it. The problem is that in Lexer 2, there is not property type declared while on Lexer 3 there is.

.github/workflows/tests.yml
egulias: If we add `lowest` to `matrix.deps` should that work for both versions?

derrabus: Yes. Do you want a deps=lowest job for both PHP versions? I felt like one of those is enough.

egulias:
Hi @derrabus!
Scrutinizer is no longer able to find code coverage. I believe that we need to change `.scrutinizer.yml` and add `file: build/logs/clover.xml` could fix it.

derrabus:
Let's try it.

.github/workflows/tests.yml
egulias: Ok, yes, one is enough.

egulias:
Nope. I will work on it, I might need to do multiple test. Will merge and fix it later... in 3.x was working.

PR #353, opened 2023-01-07
V4 readme updates

egulias:


egulias:
Scrutinizer is failing because of the container is not activating xdebug properly.

Issue #338, opened 2022-12-05
DNS record check should pass if email address has no top-level domain 


SuperDJ:
Right not the DNSCheckValidation check fails when an email address has no top-level domain.

egulias:
Hi. I don't understand. Can you clarify which is the problem?

SuperDJ:
@egulias I have noticed in your code, there are multiple checks to be made to validate an email address. One of these check if an domain has an IP-address but in some rare case an domain like "example.com" hasn't an IP but "www.example.com" has. After the first check on [#L135](https://github.com/egulias/EmailValidator/blob/3.x/src/Validation/DNSCheckValidation.php#L135) it just stops, even though the domain "example.com" has an valid MX-record.

egulias:
Thanks for clarifying.
I see, kind of a miss configuration.
I'll flag it to be added. I'm happy to review a PR from you with a solution if it's urgent for you.

kergekacsa:
Let's see if this helps. #355

egulias:
Closed by https://github.com/egulias/EmailValidator/pull/355

PR #355, opened 2023-01-09
DNS record check now passes if email address has no top-level domain

kergekacsa:
For request #338 

Ps: Sorry the previous one.


egulias:
Merged. Thanks @kergekacsa !

PR #364, opened 2023-05-17
code refactor

xHeaven:
This is a code refactor. Read the commit messages and descriptions for more information on each commit.


egulias:
Hi @xHeaven , thanks for the PR.
It does have a lot of changes, I'll be reviewing it.
Meanwhile, please review Psalm errors (you should be able to run it locally too)

```
ERROR: InvalidPropertyAssignmentValue - src/EmailLexer.php:161:28 - $this->nullToken 
with declared type 'Doctrine\Common\Lexer\Token<int, string>' 
cannot be assigned type 'Doctrine\Common\Lexer\Token<, "">' (see https://psalm.dev/145)
        $this->nullToken = $nullToken;
```
```
ERROR: InvalidPropertyAssignmentValue - src/EmailLexer.php:163:26 - $this->current 
with declared type 'Doctrine\Common\Lexer\Token<int, string>' 
cannot be assigned type 'Doctrine\Common\Lexer\Token<, "">' (see https://psalm.dev/145)
        $this->current = $this->previous = $this->nullToken;
```

```
ERROR: InvalidPropertyAssignmentValue - src/EmailLexer.php:163:44 - $this->previous 
with declared type 'Doctrine\Common\Lexer\Token<int, string>' 
cannot be assigned type 'Doctrine\Common\Lexer\Token<, "">' (see https://psalm.dev/145)
        $this->current = $this->previous = $this->nullToken;
```

Thanks!

xHeaven:
Hey @egulias, the Psalm errors should be fixed.
Apparently, even though the type hint that I've just reverted is absolutely redundant in the EmailLexer, it's needed, because Psalm can't seem to do its static analysis without it.
I suppose that was the original reason for having a dummy variable called `$nullToken` as well.
Too bad, weird stuff I guess.

xHeaven:
Is there anything I can do to help successfully execute the currently failing test? I've got 9-ish emails about it so far, let me know if there's anything I can do. @egulias

egulias:
Hey @xHeaven , apologies for the spam! I didn't realise you were getting the emails too.
The issue is that for some reason I haven't found yet, GitHub's action is not bringing the value of the secret ´CODACY_PROJECT_TOKEN´ ([here](https://github.com/egulias/EmailValidator/blob/4.x/.github/workflows/tests.yml#L54)), that it has worked few days ago when I first introduced it.

Will use this PR to test, you might get more spam, apologies in advance now :).

Still working on the review, thanks for your patience!

xHeaven:
No worries, keep me updated! :)

README.md
```diff
@@ -6,50 +6,54 @@
...
+|  v3.x   | 2020/12/29 |  -  |            X            |      |
+| v2.1.x  | 2016/05/16 | YES |                         |      |
+|  v1.2   | 2013/19/05 | YES |                         |      |
 
+## Requirements
 
-## Requirements ##
+* PHP 8.1
+* [Composer](https://getcomposer.org) is required for installation
+* [Spoofchecking](/src/Validation/Extra/SpoofCheckValidation.php)
```
egulias: Any reason for splitting this line in 3 on the readme?

README.md
```diff
@@ -6,50 +6,54 @@
...
 ```shell
 composer require egulias/email-validator
 ```
 
-## Getting Started ##
+## Getting Started
 
-`EmailValidator` requires you to decide which (or combination of them) validation/s strategy/ies you'd like to follow for each [validation](#available-validations).
+`EmailValidator` requires you to decide which (or combination of them) validation/s strategy/ies you'd like to follow
+for each [validation](#available-validations).
```
egulias: Please keep it on the previous line.

README.md
```diff
@@ -60,15 +64,19 @@ $validator = new EmailValidator();
...
+### Available validations
 
 1. [RFCValidation](/src/Validation/RFCValidation.php): Standard RFC-like email validation.
-2. [NoRFCWarningsValidation](/src/Validation/NoRFCWarningsValidation.php): RFC-like validation that will fail when warnings* are found.
-3. [DNSCheckValidation](/src/Validation/DNSCheckValidation.php): Will check if there are DNS records that signal that the server accepts emails. This does not entail that the email exists.
-5. [MultipleValidationWithAnd](/src/Validation/MultipleValidationWithAnd.php): It is a validation that operates over other validations performing a logical and (&&) over the result of each validation.
-6. [MessageIDValidation](/src/Validation/MessageIDValidation.php): Follows [RFC2822 for message-id](https://tools.ietf.org/html/rfc2822#section-3.6.4) to validate that field, that has some differences in the domain part.
-7. [Your own validation](#how-to-extend): You can extend the library behaviour by implementing your own validations.
+2. [NoRFCWarningsValidation](/src/Validation/NoRFCWarningsValidation.php): RFC-like validation that will fail when
+   warnings* are found.
```
egulias: I don't see this improving the reading. Please keep the original format. Apply this comment to all the lines below.

README.md
```diff
@@ -89,30 +97,33 @@ $multipleValidations = new MultipleValidationWithAnd([
...
 Please follow the [Contribution guide](CONTRIBUTING.md). Is short and simple and will help a lot.
 
-## Other Contributors ##
+## Other Contributors
 
 (You can find current contributors [here](https://github.com/egulias/EmailValidator/graphs/contributors))
 
 As this is a port from another library and work, here are other people related to the previous one:
 
-* Ricard Clau [@ricardclau](https://github.com/ricardclau):      	Performance against PHP built-in filter_var (v2 and earlier)
```
egulias: Please keep this list in it's original format.

README.md
```diff
@@ -89,30 +97,33 @@ $multipleValidations = new MultipleValidationWithAnd([
...
+Validations not present in the RFCs
 
+1. [SpoofCheckValidation](/src/Validation/Extra/SpoofCheckValidation.php): Will check for multi-utf-8 chars that can
+   signal an erroneous email name.
 
-### How to extend ###
+### How to extend
 
-It's easy! You just need to implement [EmailValidation](/src/Validation/EmailValidation.php) and you can use your own validation.
+It's easy! You just need to implement [EmailValidation](/src/Validation/EmailValidation.php) and you can use your own
```
egulias: Undo this change please.

src/EmailParser.php
```diff
@@ -56,7 +56,7 @@ private function processLocalPart(): Result
         $localPartParser = new LocalPart($this->lexer);
         $localPartResult = $localPartParser->parse();
         $this->localPart = $localPartParser->localPart();
-        $this->warnings = array_merge($localPartParser->getWarnings(), $this->warnings);
+        $this->warnings = [...$localPartParser->getWarnings(), ...$this->warnings];
```
egulias, line 59: Can you explain what is the benefit of this change?

src/Parser/CommentStrategy/DomainComment.php
```diff
@@ -12,11 +12,7 @@ class DomainComment implements CommentStrategy
 {
     public function exitCondition(EmailLexer $lexer, int $openedParenthesis): bool
     {
-        if (($openedParenthesis === 0 && $lexer->isNextToken(EmailLexer::S_DOT))) { // || !$internalLexer->moveNext()) {
-            return false;
-        }
-
-        return true;
+        return !($openedParenthesis === 0 && $lexer->isNextToken(EmailLexer::S_DOT));
```
egulias, line 15: 👍

src/Validation/DNSRecords.php
```diff
@@ -8,12 +8,12 @@ class DNSRecords
     /**
      * @var array $records
      */
-    private $records = [];
+    private $records;
```
egulias: Please keep it as an empty array.

src/Validation/DNSRecords.php
```diff
@@ -8,12 +8,12 @@ class DNSRecords
...
 
     /**
      * @var bool $error
      */
-    private $error = false;
+    private $error;
```
egulias: Please keep it as boolean false, rather than null.

src/Validation/MultipleValidationWithAnd.php
```diff
@@ -25,7 +25,7 @@ class MultipleValidationWithAnd implements EmailValidation
     /**
      * @var EmailValidation[]
      */
-    private $validations = [];
+    private $validations;
```
egulias: Please keep it as an empty array.

src/EmailParser.php
egulias: I'm not commenting on all the instances of similar changes to avoid generating noise.

src/Parser/Comment.php
```diff
@@ -66,29 +66,28 @@ public function parse(): Result
...
         return $finalValidations;
     }
 
 
     /**
-     * @return bool
+     * @return void
      */
-    private function warnEscaping(): bool
+    private function warnEscaping(): void
```
egulias, line 78: I like the change.
Changing this, we should find a better way for the implementation I believe, avoiding empty returns for a void method.

egulias (changes requested):
Hey. Thanks for waiting.
I have left some comments for you. Many are aesthetics more than anything, that I'd like to keep.
A couple need some conversation.

src/EmailParser.php
xHeaven: It's supposed to be a performance-related change. See [this benchmark](https://onlinephp.io/c/3038f).

src/Validation/DNSRecords.php
xHeaven: The initializer is redundant as the constructor immediately overwrites it.
We can either drop the redundant initializer or to make things more explicit, we could make this `private array $records` and drop the comment above which will also be redundant then.

src/Validation/DNSRecords.php
xHeaven: The initializer is redundant as the constructor immediately overwrites it.
We can either drop the redundant initializer or to make things more explicit, we could make this `private bool $error` and drop the comment above which will also be redundant then.

src/Validation/MultipleValidationWithAnd.php
xHeaven: The initializer is redundant as the constructor immediately overwrites it.
We can either drop the redundant initializer or to make things more explicit, we could make this `private array $validations` and drop the comment above which will also be redundant then.

README.md
xHeaven: The fact that it's recommended to have a specific max length for lines, in this case, it's 120 characters, which is the default hard-wrap setting in PHPStorm.
Knowing this, shall I undo these changes still?

src/Parser/Comment.php
xHeaven: Not sure if we have to avoid empty returns to be honest.
A `void` function is basically a function that doesn't have any kind of return value. The 2 returns in this function are only there to short-circuit the code, nothing else.
Not returning anything is perfectly fine in this scenario - especially because it's an internal method, only being used in one place without storing its return value anyway.

Thoughts?

README.md
xHeaven: Refer to https://github.com/egulias/EmailValidator/pull/364#discussion_r1212802161.

README.md
xHeaven: Refer to https://github.com/egulias/EmailValidator/pull/364#discussion_r1212802161.

README.md
xHeaven: Refer to https://github.com/egulias/EmailValidator/pull/364#discussion_r1212802161.

README.md
xHeaven: Refer to https://github.com/egulias/EmailValidator/pull/364#discussion_r1212802161.

README.md
egulias: Yes, I'm aware of line length recommendations. And I do enforce them in code.
However, the readme and documentation (from my POV) are not tied to it as most browsers and/or editors will wrap the lines to fit the reader screen size/definition. Adding breaking lines would break that adaptation I believe.

So, yes, please undo them for the readme. More of a personal choice here. Thanks

src/Validation/DNSRecords.php
egulias: Good point. I like the explicit typing. Let's go with that solution.

src/Validation/DNSRecords.php
egulias: Yep, good point. Let's go with the type.

src/Validation/MultipleValidationWithAnd.php
egulias: Type, thanks.

src/Parser/Comment.php
egulias: I agree, I'm not looking at changing to return something specific.
My thoughts are more around those 2 short-circuits and if there is a way to improve what is being done.
Would it be more clear an `if-if else-else` statement rather than 2 separate ifs?

If nothing comes up, I'm fine with it, just wanted to have your ideas on it.

xHeaven: I think the current solution is just fine, functional, and readable. My best idea would be merging the 2 ifs into one, but considering the length of the conditions, it might be better as-is. Resolving this for now, but I'm not against changes if this comes up later again.

xHeaven:
Hey @egulias,

I'm done with the requested changes, would you take a look at it whenever you can?
Also, I've used [constructor property promotion](https://php.watch/versions/8.0/constructor-property-promotion) in my [last commit](https://github.com/egulias/EmailValidator/pull/364/commits/5ca02b4334bc79756afa344372fbb6f9004c605e), is that fine, or would you like to stick with the old-style property definition?

egulias (approved):
Looks good! Will take the chance to see if I can manually send the coverage results and then will merge.
Thanks for the effort!

egulias:
Thanks again!

PR #367, opened 2023-05-31
Work around deprecations from doctrine/lexer

nicolas-grekas:
Version 3 of egulias/email-validator is the latest that works on PHP 7.2, thus it's the version that we use to test Symfony 5.4 on PHP 7.2.

Since we enabled reporting of deprecations from Doctrine, we are flooded with notices triggered by this package, because it relies on the deprecated array-style to read tokens.

This PR fixes it.

A quick merge would be appreciated to remove so many failures on our CI :pray: 

And a release would help ppl on 7.2 to not have these deprecations also of course.

### DIFF
--- a/src/EmailLexer.php
+++ b/src/EmailLexer.php
@@ -213,13 +213,11 @@ class EmailLexer extends AbstractLexer
     public function moveNext() : bool
     {
         if ($this->hasToRecord && $this->previous === self::$nullToken) {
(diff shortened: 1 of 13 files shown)


fancyweb (approved):

egulias:
Thanks for taking the time @nicolas-grekas 

On it.

nicolas-grekas:
Thanks for merging quickly :pray:

PR #368, opened 2023-06-09
chore: update phpunit 9 => 10 and psalm 4 => 5

Chris53897:
The <source> Element is new in PhpUnit 10.2.

The removed attributes in phpunit.xml.dist are the default ones.

### DIFF
--- a/tests/EmailValidator/EmailLexerTest.php
+++ b/tests/EmailValidator/EmailLexerTest.php
@@ -51,11 +51,11 @@ class EmailLexerTest extends TestCase
         $this->assertEquals(EmailLexer::INVALID, $lexer->current->type);
     }
 
-    public function invalidUTF8CharsProvider()
+    public static function invalidUTF8CharsProvider()
     {
         $chars = array();
         for ($i = 0; $i < 0x100; ++$i) {
-            $c = $this->utf8Chr($i);
+            $c = self::utf8Chr($i);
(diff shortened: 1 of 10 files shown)


tests/EmailValidator/EmailLexerTest.php
```diff
@@ -64,7 +64,7 @@ public function invalidUTF8CharsProvider()
         return $chars;
     }
 
-    protected function utf8Chr($code_point)
+    protected static function utf8Chr($code_point)
```
egulias, line 67: Can you explain the reasoning behind this change please?

egulias (commented):
Hey, thanks for the PR. 
Can you explain the reasoning behind all the changes to ´static´? 
I don't think they bring any meaningful improvement.

tests/EmailValidator/EmailLexerTest.php
Chris53897: PhpUnit 10 is deprecating data providers that are not static.
https://github.com/sebastianbergmann/phpunit/commit/9caafe2d49b33a21f87db248a8ad6ca7c7bdac09
https://docs.phpunit.de/en/10.0/writing-tests-for-phpunit.html#data-providers

This will trigger warnings in phpunit testruns. Some CI break with warnings.

Chris53897: This particular change is needed to make the data provider static, that is calling this function.

egulias:
Thanks!

PR #369, opened 2023-06-09
chore: replace scalar with more precise type specification

kesselb:


egulias:
Hey @kesselb , thanks for the PR. Can you check Psalm errors please? It is complaining about the changes you are doing.

kesselb:
Hi :wave: 

Token.type is defined as UnitEnum|string|int|null as per https://github.com/doctrine/lexer/blob/e74756f7517d72c238b9163fcd1ed54eb1f92bd0/src/Token.php#L15 and https://github.com/doctrine/lexer/blob/e74756f7517d72c238b9163fcd1ed54eb1f92bd0/src/Token.php#L33.

I see two options:

1) Add a runtime check for UnitEnum and handle the situation.
2) Don't document UnitEnum. It looks unlikely to me, that Token.type will be UnitEnum in our case. 

Updated the pull request for version 1. 
Please let me know when you prefer version 2.

egulias:
Thanks @kesselb

PR #375, opened 2023-09-06
Make private functions protected in DNSCheckValidation.php

williamdes:
I have to add some lines in `validateDnsRecords` 

```php
        // It was added in checkDns()
        $hostWithoutFinalDot = substr($host, 0, -1);
        if (in_array($hostWithoutFinalDot, self::BAD_TYPO_DOMAINS)) {
            return false;
        }
```

And this will abort the domain checking for BAD_TYPO_DOMAINS that are typo domains.
Like : `hormail.fr` typo of `hotmail.fr`
Passes MX check and all other tests.
But is a typo and then invalid

### DIFF
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -147,7 +147,7 @@ class DNSCheckValidation implements EmailValidation
      *
      * @return bool True on success.
      */
-    private function validateDnsRecords($host): bool
+    protected function validateDnsRecords($host): bool
     {
         $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, static::DNS_RECORD_TYPES_TO_CHECK);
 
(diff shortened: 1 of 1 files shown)


egulias:
The typo does not render a valid email per the RFC, invalid.
If you want to check fro typos you can extend with the [EmailValidator interface ](https://github.com/egulias/EmailValidator#how-to-extend) and add the validator to the validators to use.

williamdes:
> The typo does not render a valid email per the RFC, invalid. If you want to check fro typos you can extend with the [EmailValidator interface ](https://github.com/egulias/EmailValidator#how-to-extend) and add the validator to the validators to use.

Hello
Did you really look at my diff?
It would be nice that I could extend this class with my own validation needs without having to do code copy paste and try to say in sync with your code
What do you think?

I agree that typo validation is not RFC stuff but anyway, what can you do else when typo domains are perfectly technically valid..
Example: hormail.fr for hotmail.fr

PR #383, opened 2024-03-14
Update GitHub Actions to V4

DannyvdSluijs:
This PR bumps the references from v3 to v4 based on the information https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/#what-you-need-to-do and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions


egulias:
Thanks!

PR #406, opened 2025-02-12
Do not place newline after last/single error

LeonMelis:
Do not place a newline after the last error message. Also works if there is only one error message.

### DIFF
--- a/src/Result/MultipleErrors.php
+++ b/src/Result/MultipleErrors.php
@@ -41,12 +41,7 @@ class MultipleErrors extends InvalidEmail
 
     public function description() : string
     {
-        $description = '';
-        foreach($this->reasons as $reason) {
-            $description .= $reason->description() . PHP_EOL;
-        }
-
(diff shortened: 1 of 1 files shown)


egulias:
Please check the tests https://github.com/egulias/EmailValidator/actions/runs/13290888207/job/38343362212?pr=406

egulias:
And thanks for the PR!

LeonMelis:
I used the GitHub in-browser editing mode to create this quick 'n dirty PR, I didn't think of the tests, apologies!

It looks like I just need to adjust the expected outcome of the test. But to do so I guess I'll have to do a 'real' PR flow (fork, clone, edit, run tests, commit, PR), which probably means we'll have to close this PR and then open a new one.

Not sure when I'll have time for that though. If you don't want to wait you can also just close this PR and use my code in your own commit. I don't need the credit or recognition for it, I just wanted to help :-)

Issue #150, opened 2017-05-04
Reusing an EmailValidator instance in strict mode fails valid emails


edhgoose:
I've discovered that if you "reuse" an email validator with strict mode, then the warnings are never cleared. This means that if you get one invalid email and reuse the validator, all the subsequent emails that pass through the validator will in turn be marked as invalid.

We're using Symfony 2.8 (and therefore tied to v1.2) which doesn't reuse the validator as far as I can tell and therefore is not affected, but our own code was instantiating an emailValidator once and then reusing it to validate multiple email addresses.

This test case illustrates:

```PHP
    public function testOneInvalidFollowedByValid()
    {
        $emailThatGeneratesWarning = "1234567890123456789012345678901234567890123456789012345678901234567890@gmail.com";
        $emailWithNoWarnings = "hello@example.com";

        $emailValidator = new EmailValidator();

        $result1 = $emailValidator->isValid($emailWithNoWarnings, false, true);

        $this->assertTrue($result1, "Expected this to pass. In strict mode, it's a valid email");
        
        $result2 = $emailValidator->isValid($emailThatGeneratesWarning, false, true);

        $this->assertFalse($result2, "Expected this to fail. In strict mode, it's an invalid email");

        $result3 = $emailValidator->isValid($emailWithNoWarnings, false, true);

        // this assertion will unexpectedly fail.
        $this->assertTrue($result3, "Expected this to pass. It's the same email as result1.");
    }
```

It looks like the fix is to clear out the warnings before, maybe here: https://github.com/egulias/EmailValidator/blob/1.2/src/Egulias/EmailValidator/EmailValidator.php#L119. The EmailParser will also need something to reset the warnings within, otherwise they just come back.

I don't believe v2 is affected - but I haven't spent as much time confirming.

egulias:
Thanks for the report @edhgoose .

egulias:
Closing as v1.2 is no longer maintained. Thanks.

PR #409, opened 2025-03-04
Fix issue that local part parser uses remaining code from lexer with multiple email addresses

stefangr:
This issue resolves an issue when you use the EmailValidator for multiple email addresses without creating a new instance for every email address (Similar to [issue 150](https://github.com/egulias/EmailValidator/issues/150)).

Without this fix the added test will output

```
1) Egulias\EmailValidator\Tests\EmailValidator\EmailParserTest::testMultipleEmailAddresses Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'another-local-part'
+'some-random-but-large-domain-part.example.comanother-local-part'
```

### DIFF
--- a/src/Parser/LocalPart.php
+++ b/src/Parser/LocalPart.php
@@ -34,6 +34,7 @@ class LocalPart extends PartParser
 
     public function parse(): Result
     {
+        $this->lexer->clearRecorded();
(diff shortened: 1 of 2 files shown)


egulias:
Thanks for the PR @stefangr !

Issue #408, opened 2025-02-26
On production specific domains can not be resolved


alexander-schranz:
I have a IONOS and google cloud server and some A records fail to resolve the `DNS_A + DNS_MX` part:

```bash
php -r 'require_once("vendor/autoload.php"); $dnsCheckValidation = new \Egulias\EmailValidator\Validation\DNSCheckValidation(); var_dump($dnsCheckValidation->isValid("somemail@adif.es", new \Egulias\EmailValidator\EmailLexer())); var_dump($dnsCheckValidation->getError());'
```

Locally all works like expected.

If I change todo only a MX Lookup and not a A lookup all works like expected:

```diff
-        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A + DNS_MX);
+        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_MX);
```

So maybe change to something like:

```php
        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_MX);

        $dnsRecords = [];
        if (! $dnsRecordsResult->withError()) {
            $dnsRecords = $dnsRecordsResult->getRecords();
        }

        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A);

        if (! $dnsRecordsResult->withError()) {
            $dnsRecords = array_merge($dnsRecords, $dnsRecordsResult->getRecords());
        }

        // Combined check for A+MX+AAAA can fail with SERVFAIL, even in the presence of valid A/MX records
        $aaaaRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_AAAA);

        if (! $aaaaRecordsResult->withError()) {
            $dnsRecords = array_merge($dnsRecords, $aaaaRecordsResult->getRecords());
        }
```

Think this is similar to @crishoj https://github.com/egulias/EmailValidator/pull/376

alexander-schranz:
Not sure when we should then throw the `UnableToGetDNSRecord` exception, only if all lookup fails?


```php
        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_MX);

        $dnsRecords = [];
        if (! $dnsRecordsResult->withError()) {
            $dnsRecords = $dnsRecordsResult->getRecords();
        }

        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A);

        if (! $dnsRecordsResult->withError()) {
[...]

PR #410, opened 2025-03-06
Fix connection timeout issue when try combine MX + A records for specfic DNS when A record not exists

alexander-schranz:
Copied from https://github.com/egulias/EmailValidator/issues/408

----

I have a IONOS and google cloud server and some A records fail to resolve the `DNS_A + DNS_MX` part:

```bash
php -r 'require_once("vendor/autoload.php"); $dnsCheckValidation = new \Egulias\EmailValidator\Validation\DNSCheckValidation(); var_dump($dnsCheckValidation->isValid("somemail@adif.es", new \Egulias\EmailValidator\EmailLexer())); var_dump($dnsCheckValidation->getError());'
```

Locally all works like expected.

PS: Tried today a Hetzner Cloud server which has the same issue. I also tried directly the PHP:

```php
php -r 'var_dump(dns_get_record("adif.es", DNS_A + DNS_MX));'
```

And it fails. Did found out the used DNS and tried `dig` command which ends in:

```sh
dig @185.12.64.2 +short A adif.es
```

Goes into `;; connection timed out; no servers could be reache`

while over google it not errors just return nothing what is expected as only a `MX` record exists:

```sh
dig @8.8.8.8 +short A adif.es
```

### DIFF
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -181,6 +181,11 @@ class DNSCheckValidation implements EmailValidation
             return false;
         }
 
+        return $this->validateMxRecords($dnsRecords);
+    }
+
+    private function validateMxRecords($dnsRecords): bool
+    {
         // For each DNS record
         foreach ($dnsRecords as $dnsRecord) {
             if (!$this->validateMXRecord($dnsRecord)) {
@@ -191,6 +196,7 @@ class DNSCheckValidation implements EmailValidation
                 return false;
             }
         }
+
         return true;
     }
 
--- a/src/Validation/DNSCheckValidation.php
+++ b/src/Validation/DNSCheckValidation.php
@@ -17,7 +17,7 @@ class DNSCheckValidation implements EmailValidation
     /**
      * Reserved Top Level DNS Names (https://tools.ietf.org/html/rfc2606#section-2),
      * mDNS and private DNS Namespaces (https://tools.ietf.org/html/rfc6762#appendix-G)
-     * 
+     *
      * @var string[]
      */
     public const RESERVED_DNS_TOP_LEVEL_NAMES = [
@@ -145,14 +145,20 @@ class DNSCheckValidation implements EmailValidation
      */
     private function validateDnsRecords($host): bool
     {
-        $dnsRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A + DNS_MX);
+        $dnsRecords = [];
 
-        if ($dnsRecordsResult->withError()) {
-            $this->error = new InvalidEmail(new UnableToGetDNSRecord(), '');
-            return false;
+        $mxRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_MX);
+
+        if (! $mxRecordsResult->withError()) {
+            $dnsRecords = $mxRecordsResult->getRecords();
         }
 
-        $dnsRecords = $dnsRecordsResult->getRecords();
+        // Combined check for A+MX can fail with connection timed out, even in the presence of valid MX record
+        $aRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A);
+
+        if (! $aRecordsResult->withError()) {
+            $dnsRecords = array_merge($dnsRecords, $aRecordsResult->getRecords());
+        }
 
         // Combined check for A+MX+AAAA can fail with SERVFAIL, even in the presence of valid A/MX records
         $aaaaRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_AAAA);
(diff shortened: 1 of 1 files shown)


src/Validation/DNSCheckValidation.php
```diff
@@ -163,6 +169,14 @@ private function validateDnsRecords($host): bool
 
         // No MX, A or AAAA DNS records
         if ($dnsRecords === []) {
+            if ($mxRecordsResult->withError()
+                && $aRecordsResult->withError()
+                && $aaaaRecordsResult->withError()
+            ) {
+                $this->error = new InvalidEmail(new UnableToGetDNSRecord(), '');
+                return false;
+            }
```
alexander-schranz, line 178: unsure when we should return `UnableToGetDNSRecord` if one DNS lookup failed or when all failed?

src/Validation/DNSCheckValidation.php
```diff
@@ -177,6 +196,7 @@ private function validateDnsRecords($host): bool
                 return false;
```
alexander-schranz, line 201: is this expected as soon as one record doesn't pass validateMXRecord it marks the domain as failed? Didn't found any issue with it but thought about if we should do instead of 3 lookup check after every lookup if we already found a valid dns record and early return true. But that would mean this line would not work as expected that it early return false.

alexander-schranz: personally I would expect something like this:

```php
        // For each DNS record
        foreach ($dnsRecords as $dnsRecord) {
            if ($this->validateMXRecord($dnsRecord)) {
                return true; // do not check all domains
            }
            
            //if (empty($this->mxRecords)) {
            //    $this->warnings[NoDNSMXRecord::CODE] = new NoDNSMXRecord();
            // }
        }

        return false;
```

and then do something like todo A and AAAA check only when MX failed:

```diff
        $mxRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_MX);

        if (! $mxRecordsResult->withError()
+           && $this->validateMxRecords($mxRecordsResult->getRecords())) {
-            $dnsRecords = $mxRecordsResult->getRecords();
+            return true;
        }

        // Combined check for A+MX can fail with connection timed out, even in the presence of valid MX record
        $aRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_A);

        if (! $aRecordsResult->withError()
+           && $this->validateMxRecords($aRecordsResult->getRecords())) {
-            $dnsRecords = array_merge($dnsRecords, $aRecordsResult->getRecords());
+            return true;
        }

        // Combined check for A+MX+AAAA can fail with SERVFAIL, even in the presence of valid A/MX records
        $aaaaRecordsResult = $this->dnsGetRecord->getRecords($host, DNS_AAAA);

        if (! $aaaaRecordsResult->withError()
+           && $this->validateMxRecords($aaaaRecordsResult->getRecords())) {
-            $dnsRecords = array_merge($dnsRecords, $aaaaRecordsResult->getRecords());
+            return true;
        }
        
+        if ($mxRecordsResult->withError()
+            && $aRecordsResult->withError()
+            && $aaaaRecordsResult->withError()
+        ) {
+            $this->error = new InvalidEmail(new UnableToGetDNSRecord(), '');
+            return false;
+        }
+        if ([] === $mxRecordsResult->getRecords()
+            && [] === $aRecordsResult->getRecords()
+            && [] === $aaaaRecordsResult->getRecords()
+        ) {
+            $this->error = new InvalidEmail(new ReasonNoDNSRecord(), '');
+            return false;
+        }

+        $this->warnings[NoDNSMXRecord::CODE] = new NoDNSMXRecord();
        
        return false;
```

but I'm not deep into this Lib or DNS records and maybe the early `return false` make sense here to avoid something unexpectly happening.

egulias: The general strategy of the lib is _fail fast_ . Now for this case, given that there are network roundtrips was to "_confirm fase_" if that make sense. Hence not checking all records, with one we do have a valid MX. Happy to change if you believe will improve accuracy.

With a more in depth validation like your proposal, I agree to the general idea you are suggesting on this comment: we go from more accurate to less accurate and we short-circuit when we find a match.

Implementation wise, I'd go for something like
```php
        if (! $mxRecordsResult->withError()
            && $this->validateMxRecords($mxRecordsResult->getRecords())) {
             $dnsRecords = $mxRecordsResult->getRecords();
              return true;
        }
       // We know there are no MX records already
       $this->warnings[NoDNSMXRecord::CODE] = new NoDNSMXRecord();
      //If you feel like so, you can create new warnings for each missing DNS record
      //...

     //do the logic for error setting here
     $this->setError($mxRecordsResult, $aRecordsResult, $aaaaRecordsResult)
```

egulias (commented):
Thanks for the PR @alexander-schranz , did a comment on your thoughts. I agree with them, just added some notes.

egulias:
Please check Pslam, its complaining :)

